Issue #6, opened 2017-08-04
Write up an explanation of parametrized IDs in modular docs


asteflova:
- (Some) writers don’t understand the need for [#id-{parameters}].
- The mod docs initiative has been asked (by the AsciiDoc Mark-up Conventions init.) to provide a doc that would explain the benefits of this approach and how it works.

ritz303:
My goal is to have a draft of this content ready for review by the end of August.

ritz303:
Submitted a [pull request](https://github.com/redhat-documentation/modular-docs/pull/10) to have my rough draft reviewed by the team.

PR #10, opened 2017-08-30
Modular doc parametrized anchor tag ids

ritz303:
First rough draft for using doc attributes in an anchor id.  This allows modules to be reused in the same assembly.

I placed this concept module in the appendix examples for now.

This pull request addresses issue #6.

### DIFF
--- a/modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
+++ b/modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
@@ -2,32 +2,10 @@
 
 = Reusing a Module in an Assembly
 
-In some cases, a book will need to include the same module multiple times within the same assembly.
+With some user stories, an assembly will need to include the same module multiple times.
 How do you reuse modules within the same assembly without getting build errors for using duplicate anchor IDs?
-The answer is embedding a user-defined document attribute in the anchor ID name.
-
-Programming languages call this variable substitution, which is also know as variable interpolation.
-The AsciiDoc markup language is doing the same type of substitution with document attributes.
-
-There are three syntax formats for creating anchor IDs within a module or an assembly:
-
-* *Standard ID Syntax:*
-+
-----
-[[anchor-id]]
-----
-
-* *Short-hand ID Syntax:*
-+
-----
-[#anchor-id]
-----
-
-* *Long-hand ID Syntax:*
-+
-----
-[id='anchor-id']
-----
+By embedding a user-defined document attribute in the anchor ID name.
+This allows you to link:http://asciidoctor.org/docs/user-manual/#include-multiple[include a module multiple time] in the same assembly.
 
 [discrete]
 == Examples
@@ -58,7 +36,7 @@ Using the standard syntax format will render the appendix correctly.
 
 .Practical Example
 
-In a book about data replication, you want to reuse a procedure module for monitoring the progress of data replication in two different chapters.
+In a user story about data replication, you want to reuse a procedure module for monitoring the progress of data replication in two different chapters.
 
 The assembly file contains the following content:
 ----
--- a/modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
+++ b/modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
@@ -1,6 +1,6 @@
-[#using-a-document-attribute-in-an-anchor-id]
+[#reusing-a-module-in-an-assembly]
 
-= Using a Document Attribute in an Anchor ID
+= Reusing a Module in an Assembly
 
 In some cases, a book will need to include the same module multiple times within the same assembly.
 How do you reuse modules within the same assembly without getting build errors for using duplicate anchor IDs?
--- a/modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
+++ b/modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
@@ -83,4 +83,4 @@ The first two lines of the _procedure_monitoring-the-progress-of-data-replicatio
 [discrete]
 == Additional Resources
 
-* The link:http://asciidoctor.org/docs/user-manual/[AsciiDoc User Manual].
+* The link:http://asciidoctor.org/docs/user-manual/[Asciidoctor User Manual].
--- a/modular-docs-manual/content/topics/concept_using-a-document-attribute-in-an-anchor-id.adoc
+++ b/modular-docs-manual/content/topics/concept_using-a-document-attribute-in-an-anchor-id.adoc
@@ -1,0 +1,86 @@
+[#using-a-document-attribute-in-an-anchor-id]
+
+= Using a Document Attribute in an Anchor ID
+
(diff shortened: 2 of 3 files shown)


bhardesty:
This looks good to me. My only comment is about the title - you might want to consider using a title that's more reflective of the user goal, like, "Including a Module Multiple Times in an Assembly" or "Reusing a Module in an Assembly".

ritz303:
@bhardesty : Thanks for taking the time to review the module!  You make a good point.  I've updated the content.

modular-docs-manual/content/topics/concept_using-a-document-attribute-in-an-anchor-id.adoc
```diff
@@ -0,0 +1,86 @@
...
+----
+[id='monitoring-the-progress-of-data-replication-{chapter}']
+
+= Monitoring the Progress of Data Replication
+----
+
+[discrete]
+== Additional Resources
+
+* The link:http://asciidoctor.org/docs/user-manual/[AsciiDoc User Manual].
```
phayz: s/AsciiDoc/Asciidoctor

ritz303: @phayz : Good catch!  I've updated the source content.

modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
```diff
@@ -0,0 +1,86 @@
+[#reusing-a-module-in-an-assembly]
+
+= Reusing a Module in an Assembly
+
+In some cases, a book will need to include the same module multiple times within the same assembly.
+How do you reuse modules within the same assembly without getting build errors for using duplicate anchor IDs?
+The answer is embedding a user-defined document attribute in the anchor ID name.
+
+Programming languages call this variable substitution, which is also know as variable interpolation.
```
rkratky: known

modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
```diff
@@ -0,0 +1,86 @@
+[#reusing-a-module-in-an-assembly]
+
+= Reusing a Module in an Assembly
+
+In some cases, a book will need to include the same module multiple times within the same assembly.
```
rkratky: Could we dispense with the 'book' thing? Let's only speak in terms of modules and assemblies.

modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
```diff
@@ -0,0 +1,86 @@
...
+How do you reuse modules within the same assembly without getting build errors for using duplicate anchor IDs?
+The answer is embedding a user-defined document attribute in the anchor ID name.
```
rkratky: "The answer is" -- I suggest leaving out/reformulating to be more minimalist.

modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
rkratky: Second part of the sentence is IMO superfluous.

modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+Programming languages call this variable substitution, which is also know as variable interpolation.
+The AsciiDoc markup language is doing the same type of substitution with document attributes.
```
rkratky: Let's include a link here to the Asciidoctor manual that describes this: http://asciidoctor.org/docs/user-manual/#include-multiple

modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+There are three syntax formats for creating anchor IDs within a module or an assembly:
```
rkratky: Is this needed? If so, I suggest creating a reference module out of it. But I would rather leave it out -- no need to duplicate adoc syntax manuals.

ritz303:
@rkratky : Thanks for the review!  I've made your suggested changes.

modular-docs-manual/content/topics/concept_reusing-a-module-in-an-assembly.adoc
```diff
@@ -0,0 +1,64 @@
...
+How do you reuse modules within the same assembly without getting build errors for using duplicate anchor IDs?
+By embedding a user-defined document attribute in the anchor ID name.
+This allows you to link:http://asciidoctor.org/docs/user-manual/#include-multiple[include a module multiple time] in the same assembly.
+
+[discrete]
+== Examples
+
+Only the short-hand and long-hand ID syntax formats support document attribute substitution.
+
+.Syntax Example
```
rkratky: Let's use the standard example block like in the other examples in the guide:

```
.Example Title
====
body
====
```

modular-docs-manual/content/topics/concept_using-a-document-attribute-in-an-anchor-id.adoc
TovaCohen: Now that I understand how this works, the instructions as written make sense to me. But when I read them the first time, I struggled to understand how to implement this and it took me many tries to get it right. I would rewrite this content lto be something like the following. I'd be happy to update the asciidoc file if you want me to. 
### 
Reusing a Module in an Assembly

To use a module multiple times in an assembly, you must employ a document attribute that distinguishes each instance of the reused module. Without the distinguishing document attribute, there are build errors because the same anchor ID appears multiple times in a book.

Follow these steps:

1. In the module that is being reused, define the anchor ID in one of these formats:

[#add-eggs]
[id='add-eggs']

You cannot use this format: [[add-eggs]]

2. Append a document attribute to the anchor ID. For example, suppose the name of the document attribute is **reuse**. In the module being reused, the anchor ID would be one of these:

[#add-eggs-{reuse}]
[id='add-eggs-{reuse}]

3. In an assembly that includes the module being reused, specify a value for, for example, the reuse document attribute. Do this after the heading and before the first paragraph. For example:

[[bake-cake]]
= Bake a Cake

:reuse: cake
The introductory paragraph goes here and it might have a cross-reference to the module being reused and that reference would look something like this: see <<add-eggs-{reuse}>>. Or maybe this paragraph introduces a bulleted list of cross-references. Any one of the cross-references can look like this:

. <<add-eggs-{reuse}>>.

And then of course since this is an assembly, it would have an include statement for the module being reused. There's nothing special about the include statement. For example:

include::add-eggs.adoc[leveloffset=+1]

4. Repeat the previous step for each assembly that includes the module being reused. In each assembly, specify a different value for the reuse document attribute. For example:

[[bake-muffins]]
= Bake Muffins

:reuse: muffin
Your text starts here and it can contain a cross-reference that would look like this:

See <<add-eggs-{reuse}>>

include::add-eggs.adoc[leveloffset=+1]

PR #19, opened 2017-09-22
Add/update sections on anchor names, file names, and reusing modules

sterobin:
Summary of changes:

- Added new 3.1.4: Anchor Names and Files Names
- Moved "Reusing Modules" section from Appendix to 3.2.2 Reusing Modules in Assemblies, and rewrote the section completely.
- Updated impact cross-refs elsewhere in doc.

See the [output preview](http://file.rdu.redhat.com/~sterobin/BXMSDOC-1785/).

Side note: The reason I ditched the entire content of the existing "Reusing Modules" section and added new is not that what was there was inaccurate or bad. I instead incorporated essentially the same basic info, but expanded, explained in more detail with more concrete examples, and took a stances on a specific doc attribute ({context}) that I'm suggesting we adopt universally where possible, rather than leaving it up to everyone individually. This was the main push for making this doc update, from my understanding in our previous meetings.

I'm of course open to suggestions.


rkratky:
@sterobin, you added the `build/` directory in your commit, which is just a ccutil artifact. This stuff shouldn't be committed.

Please, remove that from the PR (as you can see, your commit now adds some 400 files -- not good). I suggest you add the `build/` directory in your `.gitignore` file to prevent this from happening in the future.

sterobin:
Oops, so sorry @rkratky! Rookie mistake. I just removed my `build` directory and re-committed. The diffs in the PR are now clean(er).

rkratky:
@sterobin, one more thing: get rid of the *.html files and the `html/` directory, too. I forgot to call those out in the previous comment.

rkratky:
We're getting there :-) There's one more HTML file: `modular-docs-manual/content/modular-doc-manual.html`.

sterobin:
:) @rkratky , you and I apparently need better things to do on a Saturday. 

The reason I've left the `modular-doc-manual.html` is that I'm trying to understand why I need to remove it, or better, why it's showing up as new at all. It makes sense why I needed to remove my `build` and `html` folders, and my `master.html` file. Those were all things that got added when I was building and previewing, etc. But the `modular-doc-manual.html` did not get added. It's part of the original structure in `master`. In fact, I noticed in making the update that this file didn't ever change to reflect my updates anyway. Clearly I don't understand how this file works.

My confusion aside, I'm happy to just remove it too, but won't that mean it will be removed when this branch gets merged to `master`, since I'd be removing the file that was pulled from `master` to begin with? I just want to make sure that by my removing it, I won't mess up `master` when I merge.

Bottom line: You sure I'm cool to remove `modular-doc-manual.html`?

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -0,0 +1,52 @@
+[#anchor-and-file-names]
+= Anchor Names and File Names
+
+To optimize modular documentation, follow these guidelines for naming module anchors and files.
+
+Anchor names:: Provide an anchor for every module so that it can be identified when reused or cross-referenced. Give the anchor name the same or a similar name to the module heading, separated by underscores, and indicate the module type (`con`, `ref`, `proc`) at the end of the name. If applicable, consider also providing a module sub-type (`create`, `edit`, `delete`) before the module type:
```
tradej: I'd say the sub-types are rather arbitrary in this case, and are not defined anywhere else. I'd advocate for either codifying them somehow (which means coming up with a list of sub-types that fits a reasonable number of use cases), or removing this part altogether.

My rationale is that this suggestion is a deviation from the rule _the filename should stay as close to the module title as possible_ without clearly formulated benefits. If we don't have a list of sub-types, this becomes just a random word tacked just before the end of the filename.

tradej: I would appreciate elaborating in the parentheses about the module types, but that's just a personal preference:
```
...indicate the module type (`con` for concept, `ref` for reference, and `proc` for procedure modules) at the end...
```

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -0,0 +1,52 @@
+[#anchor-and-file-names]
+= Anchor Names and File Names
+
+To optimize modular documentation, follow these guidelines for naming module anchors and files.
```
tradej: The sentence should end with a colon [Chicago Manual of Style, 6.124].

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -0,0 +1,52 @@
...
+= Creating Guided Decision Tables
+
+You can use guided decision tables to ...
+----
+
+NOTE: Although you can use other formats for naming anchors, the format defined here is recommended and in some cases required to enable topics to be properly cross-referenced and reused. For details, see xref:reusing-modules[].
+
+For more information about Asciidoc anchors, see the link:http://asciidoctor.org/docs/user-manual/#anchordef[Asciidoctor User Manual].
+
+File names:: Give an informative name separated by dashes and indicate the module type (`con`, `ref`, `proc`) at the end of the name. If applicable, consider also providing a module sub-type (`create`, `edit`, `delete`) right before the module type. Indicating the module type and sub-type at the end of the file name instead of the beginning of the file name results in topics being sorted based on subject matter rather than module type. This assortment can be helpful for managing related content.
```
tradej: I disagree with this practice. I believe the file names should match the module title as closely as possible to facilitate finding which file is necessary to modify based on rendered HTML alone. If we stray from that, we risk always having to open `master.adoc` first and then arduously find which module the given section is rendered from.

If we want to allow differing file names and module titles, I suggest we allow discrepancies only at the end of the file names to keep the searchability of modules (most search functionality starts at the beginning of the file names).

tradej: > Indicating the module type and sub-type at the end of the file name instead of the beginning of the file name results in topics being sorted based on subject matter rather than module type.

Not exactly true if you keep anchor IDs close to module titles, especially in procedures (they usually start with a gerund).

modular-docs-manual/content/topics/module_reusing-modules-procedure.adoc
```diff
@@ -0,0 +1,237 @@
...
+[source]
+----
+For details, see xref:projects_con_intro[].
+
+or
+
+For details, see xref:projects_con_definitions[].
+----
+
+== Additional Resources
```
tradej: In the `assembly_writing-mod-docs.adoc`, this section is immediately followed by the assembly's _Additional Resources_ section, which looks strange. Can this information be put somewhere else?

tradej:
In addition to my comments here, I have also filed #21. All my suggestions are open to discussion, of course.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: Done

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: Agreed. Done.

sterobin: You're right. I see your point here, with anchor names. I don't necessarily agree, however, with your argument for file names, i.e., that the file name too should match the module title as closely as possible. I've replied accordingly in your comments that follow.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: Say you have two procedures for columns in guided decision tables: one for adding columns, one for removing columns. With your proposal, these file names would be called "adding-guided-decision-table-columns-proc.adoc" and "removing-guided-decision-table-columns-proc.adoc" (I think, if I've understood you). These would then be very far apart from each other in the project's list of alphabetized of topics, and I actually don't see how their matching the HTML output would necessarily be more helpful. Certainly could be if it's an exact match and you're basing your changes on HTML output, but if you're not looking at the HTML and you just see in a JIRA that you need to change everything in the "Guided Decision Table Columns" chapter, then you'd have to go section-by-section in the HTML to confirm the name of each to then go look for the matching modules, scattered around in the adoc file list.

With what I'm saying, these would instead be named "guided-decision-table-columns-add-proc.adoc" and "guided-decision-table-columns-remove-proc.adoc", and true, they don't look anything like the sections in the output, BUT if I want to update anything at all dealing with guided decision tables, regardless or module type, sub-type, or whatever, I can go directly to the set of "guided decision tables" files clumped together in the adoc file list and find the right module type within that set. It's a trade-off, of course, from the alternative to match the output.

I care less about the exact formatting, and thus am happy to nix or change what I've said about sub-types, etc., but my main argument is to keep the general subject matter as the unvarying identifiable content at the start of the file name so that the adoc list is a collection of clumped, related files, not a scattered array of disparate files.

I've done both what you're saying and what I'm proposing, and have personally found what I've proposed much easier and more logical when making updates.

This is only my opinion, and I'm certainly open to your and the group's suggestions, based on everyone's experience. But just wanted to counter to make sure my rationale was clear. Apologies if I didn't understand you correctly.

sterobin: See my next comment, in reply to yours.

tradej: You understood me perfectly, and I agree that the sorting of the files in the directory suffers with this. To me, it is the price to pay for being able to find any file based on the title alone. I do that quite often, but I'd like other team members to weigh in on this with their experience.

A possible solution dawned on me: What if we keep the file names close to the titles, but arrange the files into directories based on subject matter? See example:
```
topics/
  guided-decision-tables/
    adding-guided-decision-table-columns-proc.adoc
    removing-guided-decision-table-columns-proc.adoc
```

I believe this would work pretty well, at least in my head, and I think it satisfies both our requirements.

tradej: Based on our IRC discussion with @sterobin, I withdraw this suggestion, and will file a follow-up issue where we can resolve this matter somehow. In the context of this PR, consider all my concerns addressed.

modular-docs-manual/content/topics/module_reusing-modules-procedure.adoc
sterobin: I'll just remove that section. I see little value in it, IMO, at least not as much as the Additional Resources stacked right before it. Plus it's already referred to in guidelines-assemblies.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -0,0 +1,52 @@
+[#anchor-and-file-names]
```
sterobin: Apparently the discussion between @tradej and me on anchor names and file names was removed when I pushed the updated branch. I thought it would be retained in the previous commit, but I guess not.

At any rate, at the risk of throwing out a bombshell, I've actually re-proposed for anchors and files to actually be named the same. @tradej had rationale for naming anchors the same as the module heading, as we do traditionally (and which I previously said, in contradiction to myself), but I believe that now that each file will or should consist of a single anchor typically, it might make more sense for the file and anchor to match, with the exception of underscores vs. dashes (and that distinction has merit as well, which I can explain). This seems clearer to me when we cross-reference, etc. Previously, before modules, having anchor names that matched module headings made more sense because we had many anchored headings in a single file. But with a single anchor per file, for the most part, it doesn't make much sense to me to have two different naming conventions anymore. 

Bottom line, whatever we decide for the file names, we should probably apply it also to the anchor names for simplicity. Whether we agree with what I'm proposing or not is valuable discussion we need to have, but I do think they should be the same for both, either way. 

I'd like to hear more about @tradej 's reasons for the contrary.

tradej:
Based on the latest developments, I must rescind my green light for the PR. I have talked to @rkratky, and we have agreed that these are the necessary criteria for us to approve the pull request:

1) The anchor IDs should match the module heading in the following way: all letters lowercase, all non-alphanumeric characters replaced with underscores, contains the `_{context}` suffix. The reasoning is that it is not us who is the primary consumer of the anchor ID, it is the reader/user. Consider the following example (module title is _Creating Guided Decision Tables_):

    ```
    https://access.redhat.com/<snip>/guide_name.html#guided_decision_tables_create_proc

    vs

    https://access.redhat.com/<snip>/guide_name.html#creating_guided_decision_tables
    ```
    The latter makes is clearer to the reader what they will see when clicking the link, and on the off-chance that the anchor ID changed and the link therefore points to the beginning of the page, it allows the the reader to find the correct section manually (Ctrl + F, or looking at the ToC).
2) The issue of file names must be resolved in either of the following ways:
    1) The files are named exactly like the anchor ID minus the `_{context}` suffix if applicable, and they have the module type prefix or suffix. Both myself and @rkratky are heavily in favour of a prefix because you get ordering by module type for free (for sorting by subject matter, you can use directories). Example:

        ```
        proc-creating_guided_decision_tables.adoc:

        [#creating_guided_decision_tables_{context}]
        = Creating Guided Decision Tables
        ...
        ```
    2) The issue of file naming is completely removed from this PR, retaining the status quo. A follow-up issue is filed, where we discuss options of different file naming. This, of course, means the anchor ID and file names diverge, at least until the follow-up issue is resolved.
3) The concept of module sub-types is completely removed from this PR. This concept is being introduced to the initiative in this PR, and as proposed, can only be implemented in one of these ways:
    1) It is a free-form field. This, in my opinion, defeats the purpose of codifying such a field, and is only a means to arbitrary deviation of the anchor ID from the module title.
    2) There is an enumeration of admissible sub-types. This inevitably leads to some modules not falling into any of the sub-types, or falling into the _other_ sub-type, again not adding value.

Please adjust the pull request so that it complies with above requests.

tradej:
@asteflova, what do you think about this proposal?

sterobin:
@tradej , I'm not sure how I can update the PR to comply with your 3 requests, because in my mind, requests 1 and 3 (anchor names and module sub-types) hinge on request 2 (file names), which you've said is being taken up separately for discussion. If we go with your proposal on file naming (ex: proc-creating_guided_decision_tables), then there's not need for module sub-types, which only applies with my proposed naming conventions that are based on subject matter (ex: guided_decision_tables_create_proc). And I do see your point on anchor names in URLs, etc., and agree.

We all agree that we would rather not diverge anchor names and file names, since with modules there is essentially only one anchor per file, so having differing names seems unnecessary. We just don't agree on what that should be.

Bottom line, can some please answer me this: 

What value is achieved by naming files by module, such as `proc-creating_guided_decision_tables`? That is, why is it helpful to have files listed by module type? I've made my case for why it's helpful to have it listed by subject matter, but I don't think I've yet heard a counter argument on this. That, to me, determines everything else. 

@tradej and @rkratky ?

rkratky:
@sterobin,

> What value is achieved by naming files by module, such as proc-creating_guided_decision_tables? That is, why is it helpful to have files listed by module type?

I have found it a lot easier to navigate directories of modules when the modules are grouped by module type. That said, even though I would prefer the format put forward in Tomas' comment, it's the least of my worries.

I agree with Tomas that points 2 and 3 are far more important.

To sum up:

- It's important to keep file names, IDs, and module headings consistent.
- To achieve that, the easiest way is to have them be *the same* (of course, with file names and IDs in lowercase and using underscores).
- Adding module type to file names seems like good practice, and I would argue for including it as a prefix, but I could live with it being a suffix or even missing entirely.

tradej:
@sterobin,

> I'm not sure how I can update the PR to comply with your 3 requests, because in my mind, requests 1 and 3 (anchor names and module sub-types) hinge on request 2 (file names), which you've said is being taken up separately for discussion.

Not entirely. Let me clarify:

* The first point only asks that the anchor ID match the module heading, regardless of the file name. This is meant to allow for the file naming discussion to happen later.
* The rationale behind the third point is to defer the discussion about sub-types to a different PR or issue too. It may very well be reintroduced at a later time if there is a consensus we want to do such a thing, but for the time being, it raises questions that block this particular PR from proceeding (exactly because it ties these changes to the file naming issue).

> What value is achieved by naming files by module, such as proc-creating_guided_decision_tables? That is, why is it helpful to have files listed by module type?

This is the discussion I want to have out of this PR so that it doesn't block its progress. See #25.

Does the above answer your questions?

sterobin:
@rkratky and @tradej , I've left my reply in #25 .

asteflova:
> @asteflova, what do you think about this proposal?

You all seem to have though about this much more thoroughly than me so I don't want to throw my own uneducated opinion into the conversation :)

The only thing I feel strongly about is that we introduce the rules as guidelines or recommendations, not requirements. Consider how strong your opinions on this matter are; we can only expect at least some other writers to have preferences of their own or to want to experiment with what works best for them. Strict rules and processes can lead to a lot of discontent, especially if writers don't even agree with them.

sterobin:
@rkratky and @tradej, I've updated the PR per our discussions. The diffs can now be seen. The [output preview](http://file.rdu.redhat.com/~sterobin/BXMSDOC-1785/) (also linked above) is also now up to date.

Let me know what you think.

tradej:
@sterobin, thank you for the work, it is immense. I fundamentally agree with the current version, let me just update #21 with your latest changes, and add some more that I think are necessary (updating the templates with the guidelines you put forth here, etc.). It should be ready today or tomorrow. My changes do not alter the guidelines, they're just touch-ups.

@rkratky, can you please merge #21 to `Module-Reuse-Edits` once I have it ready and @sterobin agrees to my suggestions?

As for when to merge this PR, I suggest we hold off for when Aron is back. Bear with me: I am very glad we came to an agreement about the final guidelines we want to merge, and it took us quite a discussion and it was a pretty heated debate, at least in my opinion. For that reason, I think Aron should see this before it's merged. Also, we introduce some changes that affect many existing docs, namely changing the anchor ID delimiter from an underscore to a dash, so that also warrants another review by an interested party, IMO.

tradej:
@sterobin @rkratky Can you please have a look at #26? I have made some more changes. Thank you.

sterobin:
@rkratky or @tradej, would one of you mind addressing @bhardesty's comment in the conflicting file listed in this PR? It's `module_guidelines-concept.adoc`. Need to make a call to resolve the conflict, and I'm probably not the one to make it.

tradej:
@sterobin I have looked at the conflict; it's not very difficult to resolve. If you feel like trying to resolve it, it would be a good exercise IMO (plus it would be a good test whether I have documented it well enough in the BxMS Contrib Guide :) ). However, if you don't have the time or don't feel like it, I will do it (I need to create a new PR like #21 and #26 because I don't have the necessary rights to this repository AFAICT).

sterobin:
@tradej, I just meant that I didn't want to step on anyone's toes since the conflict came from elsewhere. But I went ahead and resolved the conflict, but I wanted to try using the UI conflict resolver (click Resolve Conflict and resolve there) since it was so small. Looks like it worked but also brought in whatever other changes were associated with that commit. The commit that was generated when I resolved it is called "Merge branch 'master' into Module-Reuse-Edits", which sounds a little iffy, but looks to be just the changes from whoever made that commit. I suppose this is fine and is one way to resolve it. If you see issues with this method, I'm happy to hard reset back to the previous commit and resolve the conflict the long way. Let me know.

modular-docs-manual/content/topics/module_creating-modules.adoc
```diff
@@ -10,6 +10,8 @@ After identifying and creating the assemblies, each assembly should identify the
 +
 Each assembly should already have the names of the modules that should be included in the assembly. Now you just need to create the actual files for those modules.
 
+Be sure to follow the conventions for xref:anchor-and-file-names[naming anchors and files]. For example: `guided-decision-tables-con.adoc`
```
tradej: The example is wrong here, should be `con_guided-decision-tables.adoc`. I've probably missed this before. Otherwise LGTM.

tradej:
@sterobin Apart from a [small error I found](https://github.com/redhat-documentation/modular-docs/pull/19/files#r144018182), this looks good now, thanks.

In the future, I would recommend not merging the `master` branch into this one, though, but rebasing against `master` instead. How to do that:

```
$ git checkout Module-Reuse-Edits
$ git fetch origin
$ git rebase origin/master
<RESOLVE CONFLICTS IF ANY>
$ git rebase --continue             # in case of conflicts
$ git push --force
```

Mind you that rebasing may be arduous with 13 commits because each of them may introduce new conflicts, so it's always a good idea to squash your commits to as few as possible. Here's a nice [git squash tutorial](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) in case you haven't done this before. Always make a backup branch, though.

modular-docs-manual/content/topics/module_creating-modules.adoc
sterobin: Great catch, @tradej! Fixed. Thanks.

ritz303:
There was a lot to digest here, but I got the gist of it and like these suggestions.  ACK.  Merging this PR.

Big thanks to sterobin and tradej for all their passion, thoughts and hard work on this!

PR #26, opened 2017-10-10
Changes to templates and other files re #19

tradej:
I have made some further changes to #19 that I think should be incorporated:

* The templates now contain the reusable IDs and the comment section. I think it is valuable to promote using `:context` even where not strictly necessary because upon reuse, the ID of the module does not change, therefore we don't break existing links.
* I have changed the IDs in the existing source files to incorporate the new guidelines.

Questions:

* Should I also rename the templates? They use underscores instead of dashes now. I'm afraid I could break some external links if there are any, though.

### DIFF
--- a/files/TEMPLATE_ASSEMBLY.adoc
+++ b/files/TEMPLATE_ASSEMBLY.adoc
@@ -1,7 +1,12 @@
+// Module included in the following assemblies:
+//
+// <List assemblies here, each on a new line>
+
+
 // Include an 'ID' that corresponds to the title of the assembly
 // The ID will be used as an anchor for linking to the title
 // Do not change the ID to make sure existing links keep working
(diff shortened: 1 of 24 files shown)


rkratky:
> Should I also rename the templates? They use underscores instead of dashes now. I'm afraid I could break some external links if there are any, though.

Yes. Better to break them now than later.

tradej:
OK. I will do that after lunch.

tradej:
Renamed. The PR is ready for review and merging, unless you have further comments or requests.

sterobin:
@tradej, I assume you don't need my signoff at this point for this, but in case you do and I don't realize, the anchor name changes, etc., look good to me. You're good to merge with #19 if @rkratky is good with it.

PR #30, opened 2017-11-07
Updated anchor formatting in all files and updated doc instructions accordingly.

sterobin:
Updated all anchors to format **[id='anchor-name']** (instead of **[#anchor…name]**) and updated doc instructions on Anchor Names and Reusing Modules in Assemblies with the same.

### DIFF
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -1,14 +1,14 @@
-[#anchor-and-file-names]
+[id='anchor-and-file-names']
 = Anchor Names and File Names
 
 To optimize modular documentation, follow these guidelines for naming module anchors and files:
 
-Anchor names:: Provide an anchor in the format `+++[#anchor-name]+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. Give the anchor the same or similar name as the module heading, separated by dashes:
+Anchor names:: Provide an anchor in the format `+++[id='anchor-name']+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. Give the anchor the same or similar name as the module heading, separated by dashes:
 +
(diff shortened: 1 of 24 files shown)


ritz303:
ACK.  Looks good.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -26,13 +26,17 @@ The guided decision tables feature works similarly to ...
...
-[#creating-guided-decision-tables]
+[id='creating-guided-decision-tables']
 = Creating Guided Decision Tables
 
 You can use guided decision tables to ...
 ----
 
-NOTE: The format defined here is recommended because it supports attributes that enable topics to be reused and cross-referenced properly. For details, see xref:reusing-modules[]. You can also use the format `+++[id='anchor-name']+++`, but be consistent in your choice of anchor formatting.
+[NOTE]
+.Note on Other Anchor Formats (Not Recommended)
```
laubai: Unrelated to asciidoc style specifically, but we've usually not altered the default heading of an admonition because it created work for localization (default DocBook headings were already translated in the spec). [This](https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/language-support.adoc) seems to suggest that the same is not true for asciidoc so this should be fine, but worth verifying.

laubai:
One comment re. titling admonitions, but looks good other than that.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: Wow @laubai, way to dig deep on that one! :) I personally think the value that custom block headings add in certain cases justifies the work they would require for localization. I do avoid them most of the time, but they're hard to resist on occasions where they optimize scannability and usability so much. Excellent food for thought, though.

PR #31, opened 2017-11-13
Fix template links

asteflova:
Two commits to fix what looks like a minor regression:

* Broken links (templates are available at github, not gitlab)
* A superfluous paragraph (that was moved to a different module in a previous pull request).


rkratky:
ack

ritz303:
ACK

PR #41, opened 2017-12-06
Update module and assembly examples (based on survey results)

asteflova:
The commit in this PR removes old examples and replaces them with links to new examples.


rkratky:
ack

ritz303:
ACK

Issue #15, opened 2017-09-04
Review the template for assemblies


asteflova:
Current template: https://raw.githubusercontent.com/redhat-documentation/modular-docs/master/files/TEMPLATE_ASSEMBLY.adoc

This expects that assemblies are basically just a metaprocedure, not a collection of various modules based on user stories. There are also other inaccuracies that we should look at.

PR #42, opened 2017-12-14
Review the assembly template for consistency with guidelines

asteflova:
Resolves https://github.com/redhat-documentation/modular-docs/issues/15

The new template corresponds to the recently reviewed guidelines in the Mod Docs guide and contains include statements for including modules.

### DIFF
--- a/modular-docs-manual/files/TEMPLATE_ASSEMBLY.adoc
+++ b/modular-docs-manual/files/TEMPLATE_ASSEMBLY.adoc
@@ -1,52 +1,48 @@
-// Module included in the following assemblies:
+// This assembly is included in the following assemblies:
 //
 // <List assemblies here, each on a new line>
 
+// For more details on writing an assembly, see section "Forming Assemblies" in the Modular Documentation Reference Guide.
 
-// Include an 'ID' that corresponds to the title of the assembly
-// The ID will be used as an anchor for linking to the title
-// Do not change the ID to make sure existing links keep working
-[#doing-several-tasks-in-sequence_{context}]
-= Doing Several Tasks in Sequence
+// Include an 'ID' that corresponds to the title of the assembly.
+// The ID will be used as an anchor for linking to the title.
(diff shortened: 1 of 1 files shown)


modular-docs-manual/files/TEMPLATE_ASSEMBLY.adoc
```diff
@@ -1,52 +1,48 @@
...
 
-. Link to 3rd task to perform.
+include::modules/path-to-another-module-file.adoc[leveloffset=+1]
 
-. Link to the Nth task to perform.
+include::path-to-an-assembly-file.adoc[leveloffset=+1]
 
-[discrete]
+[id='additional-resources-{context}']
 == Additional Resources
```
bhardesty: My only concern here is the placement of the assembly's "Additional Resources". If the last module in the assembly has additional resources, the presentation in the doc will be two Additional Resources sections (the final part of the module, followed immediately by the final part of the assembly). Also, if the assembly has a lot of modules, a reader could easily misinterpret what the assembly's Additional Resources pertains to (since the section would be so far apart from the assembly title and prereqs.

With that said, it is semantically correct to include the "Additional Resources" section just as you've presented it here, so I'm not sure that we should change anything.

asteflova: Agreed, this can be a bit risky. The module template include [discrete] headings for Additional Resources, while the assembly template doesn't -- so the assembly's resources will be displayed as a standard section with a numbered heading. I hope that helps at least a little bit -- the two sections for resources won't look the same.

Some time ago, I played with the idea of including the assembly's additional resources somewhere at the beginning of the assembly instead of at the end -- but that would probably introduce a whole new set of problems :)

Anyway, I think that until we have an alternative solution, we can leave this part of the template as it is.

ritz303:
Sorry, for the extremely delay here, I was on PTO.  ACK.

Issue #11, opened 2017-09-01
Decide what subheadings to use in the procedure module and update the template


baruza:
This issue includes several questions, I believe that for most of them we already came up with an answer, though. This issue should record the discussion and decisions in an open manner and track what changes should we done in the guide.

1) What subheading to use for the prerequisites section. I believe that we decided to use "Prerequisites". Any objections?

2) What subheading to use for the procedure itself. I think (from what I seen, that is is "Procedure") + is it a good idea to repeat the name of the procedure module in that title (for example Procedure: Changing the User Name)? Cons: duplication, Pros: possible better results when in searching results? Can someone who is in contact with JP Shermman check? (Nacia, Betty?)

3) What subheading to use for the additional details sections? I believe we decided to use "Additional Resources". Any objections?

4) Should we use the [discrete] tag for the subheadings so they don't show up in TOC? I believe that this is what we are already doing, but just want to confirm.

In addition, the guide should state why whatever we decide is recommended to follow (be consistent, cross-product reuse of modules, etc).

asteflova:
1. Ack
2. Procedure as the subheading - ack.
Repeating the name of the procedure module - I'd prefer not to do that. Duplication is a big con for me. The Procedure subheading is typically very close to the module heading (often separated only by a short procedure intro), so I don't think repeating it would add much value.
3. Ack
4. Ack

asteflova:
I sent https://github.com/redhat-documentation/modular-docs/pull/45 to propose a solution (based on a few conversations I took part in). The request was accepted, so I think we can close this issue.

Issue #8, opened 2017-08-22
Clarify the requirement to not change existing IDs


asteflova:
The templates currently say this:

// Do not change the ID to make sure existing links keep working

This confused some people because this sentence can be interpreted as a requirement to keep the template ID: [#doing_one_task]

asteflova:
Resolved in https://github.com/redhat-documentation/modular-docs/pull/45

PR #45, opened 2018-01-15
Update templates

asteflova:
This update fixes some inconsistencies between module templates and the recently updated writing guidelines in the Mod Docs Guide. For example:

* The template for concept modules now includes Additional Resources
* The IDs in all modules now use the syntax [id='bla-bla_{context}']

To avoid duplicating the Mod Docs Guide too much, the modules include only the most relevant writing guidelines, and link back to the Guide for more details.

The templates now also include quite a few comments to explain their contents, for example the purpose of {context}.

Most of the changes affected module templates, but I also made a few changes to the assembly template where I thought some consistency with the module templates would be nice.

This pull request resolves issues https://github.com/redhat-documentation/modular-docs/issues/11 and https://github.com/redhat-documentation/modular-docs/issues/8, but is not limited just to these issues.


ritz303:
@asteflova : Looks good! ACK.

Issue #46, opened 2018-02-12
Reference Template ID format wrong


ncbaratta:
In the reference template, we still have : 

[#reference-material_{context}]

instead of

[id='reference-material_{context}']

ritz303:
@ncbaratta : Good catch!  Thanks for fixing it!  Your PR for this issue has been merge into master, so I'm closing this issue.

PR #47, opened 2018-02-12
Issue #46 : Update ID tag to match other templates

ncbaratta:


asteflova:
Ack, thanks, Nicole.

ritz303:
@ncbaratta : Thanks! ACK.

PR #52, opened 2018-04-25
Saving and restoring :context: in assemblies to enable nesting

mrksu:
This PR adds a module describing how to save and restore the :`context:` variable in assemblies so that they can be nested with no build errors.

The changes are:

- A new module
- An include of this module
- Two diagrams illustrating the problem
- Updated template files to include the context-restoring statements


bhardesty:
This looks like a reasonable approach assuming that the scenario underlying this issue is valid.

Perhaps it's a bit out of scope to raise this within the scope of this PR, but why exactly are we mandating manually-overriding the anchor ID of the standard sections within an assembly (i.e. "Prerequisites" or "Additional Resources")? Manually overriding the Asciidoctor automatically-generated ID of every one of these sections within an assembly is (in my opinion) overkill. In the first place, you don't even need the ID unless you need to create a link to the section (which would be extremely rare). Second of all, it leads to issues like the one that this PR solves, which adds additional overhead and complication for the writer.

One possible alternative might be to not create IDs for the Prerequisites and Additional Resources sections in an assembly. Asciidoctor will generate unique IDs for these sections automatically, even for assemblies included within other assemblies. In the rare event that you need to link to one of these sections, you could either use the automatically-generated anchor ID (which, again, is already guaranteed to be unique within the scope of the document that Asciidoctor is compiling), or manually create a unique ID (perhaps using :context: as needed).

VladimirSlavik:
The issue is not so much with Additional Resources etc., but with modules included after a sub-assembly, such as:

```
assembly
├── module A
├── sub-assembly
│   ├── module B
│   ├── module C
│   └── module D
└── module E
```

In this situation, the sub-assembly sets the global variable :context: and after the "return up the tree" to include module E, its context is kept.

An alternative solution is to set context again after including an assembly within an assembly, such as:

```
= Example
:context: example

include::moduleA[]

include::assembly[]
:context: example

include::moduleE[]
```

This does not require everyone modification of all assemblies.

ritz303:
@mrksu : Sorry for the long delay in reviewing this PR.  This looks good to me, ACK.

mrksu:
No problem, thanks for merging.

I've left a comment about a related problem that I found after creating this PR in issue#54.

PR #61, opened 2018-07-24
Resolves #25

### LINKED ISSUE #25: Optimal File Naming (Follow-up to #19)
@sterobin writes in #19:

> What value is achieved by naming files by module, such as proc-creating_guided_decision_tables? That is, why is it helpful to have files listed by module type?

At the risk of misinterpreting what you said in, let me present the following argument (in case I indeed misinterpreted what you said, feel free to correct me):

* You write that you do share our view that anchor IDs should not diverge from file names very much.
* You also write that you agree with our view that anchor IDs should not diverge from the module heading very much because of the URLs.

The above necessarily means that, by transition, you agree that file names should not diverge very much from the module title. From this, I infer the following:

If file names should not diverge much from module headings—which provide no meaningful sorting capability—and we need to add a piece of text to it (the module type), we can either:

* make it a suffix, not gaining any sorting capability, or
* make it a prefix, gaining a not particularly helpful, but still non-trivial sorting capability.

As a side note, I consider prepending subject matter (again, a free-form string) a considerable deviation from the module heading.

cc @rkratky @asteflova

vikram-redhat:
Turns out the modular docs guide already had the verbiage about using the file name same as the anchors and titles. I have just deleted the bit about adding con, proc, ref and assembly from the examples etc.

I have also added that this is a recommendation, not a standard.

@tradej @sterobin @ritz303 @rkratky - PTAL.

### DIFF
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -41,15 +41,20 @@ The format defined here is recommended because it is the most stable and versati
 For more information about Asciidoc anchors, see the link:http://asciidoctor.org/docs/user-manual/#anchordef[Asciidoctor User Manual].
 --
 
-File names:: Give the module file the same name as the anchor used in it (which is the same as or similar to the module heading), also separated by dashes. Add a prefix with an underscore to the file name to indicate the module type in the format `prefix_file-name`. Use `con_` for concept, `ref_` for reference, `proc_` for procedure, and `assembly_` for assembly.
+File names:: Give the module file the same name as the anchor used in it (which is the same as or similar to the module heading). Assembly and module file names should accurately and closely reflect the title of the assembly or module.
++
+[NOTE]
+====
+These guidelines for file names are _recommended_. Project teams may adapt or deviate from these guidelines if need be.
+====
 +
 .Examples
-* `con_guided-decision-tables.adoc`  (Concept module)
-* `proc_creating-guided-decision-tables.adoc`  (Procedure module for creating)
-* `proc_editing-guided-decision-tables.adoc`  (Procedure module for editing)
-* `ref_guided-decision-table-examples.adoc`  (Reference module with examples)
-* `ref_guided-decision-table-columns.adoc`  (Reference module with column types)
-* `assembly_guided-decision-tables.adoc`  (Assembly of guided decision table modules)
+* `guided-decision-tables.adoc`  (Concept module)
+* `creating-guided-decision-tables.adoc`  (Procedure module for creating)
+* `editing-guided-decision-tables.adoc`  (Procedure module for editing)
+* `guided-decision-table-examples.adoc`  (Reference module with examples)
+* `guided-decision-table-columns.adoc`  (Reference module with column types)
+* `guided-decision-tables.adoc`  (Assembly of guided decision table modules)
 
 .Additional Resources
 


modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -41,15 +41,20 @@ The format defined here is recommended because it is the most stable and versati
 For more information about Asciidoc anchors, see the link:http://asciidoctor.org/docs/user-manual/#anchordef[Asciidoctor User Manual].
 --
 
-File names:: Give the module file the same name as the anchor used in it (which is the same as or similar to the module heading), also separated by dashes. Add a prefix with an underscore to the file name to indicate the module type in the format `prefix_file-name`. Use `con_` for concept, `ref_` for reference, `proc_` for procedure, and `assembly_` for assembly.
```
sterobin: @vikram-redhat, @tradej, @rkratky, since we're leaving that note in, which makes it clear that these file naming suggestions are strictly recommendations only, I think we should at least mention that adding module affixes is a consideration, and leave the example file names as they were. Then teams can just disregard if they like. Not requiring it is good, but not even mentioning it would be a pity for those who could get value from it but haven't thought of it. I know @tradej, @rkratky, and my teams have found great value in it, although others may not.

If we do go with this change as is, then these file name examples really aren't helpful, and in fact are troublesome if you look at the concept and assembly files, which are now identical - this will not be permitted by any authoring tool.

I'd propose the following for this section:

```
File names:: Give the module file the same name as the anchor used in it (which is the same as or similar to the module heading). Assembly and module file names should accurately and closely reflect the title of the assembly or module.
+
Consider adding affixes to file names to distinguish module types, if needed. This may improve the organization and navigability of your files depending on the structure and needs of your project. For example, you might add a prefix to the file name to indicate the module type in the format `prefix_file-name`, where `con_` is for concept, `ref_` is for reference, `proc_` is for procedure, and `assembly_` is for assembly.
+
.Examples
* `con_guided-decision-tables.adoc`  (Concept module)
* `proc_creating-guided-decision-tables.adoc`  (Procedure module for creating)
* `proc_editing-guided-decision-tables.adoc`  (Procedure module for editing)
* `ref_guided-decision-table-examples.adoc`  (Reference module with examples)
* `ref_guided-decision-table-columns.adoc`  (Reference module with column types)
* `assembly_guided-decision-tables.adoc`  (Assembly of guided decision table modules)

[NOTE]
====
These guidelines for file names are _recommended_. Project teams may adapt or deviate from these guidelines as needed.
====
```

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -41,15 +41,20 @@ The format defined here is recommended because it is the most stable and versati
...
+File names:: Give the module file the same name as the anchor used in it (which is the same as or similar to the module heading). Assembly and module file names should accurately and closely reflect the title of the assembly or module.
++
+[NOTE]
+====
+These guidelines for file names are _recommended_. Project teams may adapt or deviate from these guidelines if need be.
```
kalexand-rh: @vikram-redhat, I think your note's a bit confusing. It makes it sound like making the file name include the anchor is recommended, and I'm not sure that's your intent.

 I'd move the note under the examples and make the text something like:

"While the module file name must include the anchor name, you can annotate the file names to differentiate the modules types. If you add annotations, ensure that all members of your team use a consistent naming style.

For example, you can a prefix with an underscore to the file name to indicate the module type in the format \`prefix_file-name\`. Use \`con_\` for concept, \`ref_\` for reference, \`proc_\` for procedure, and \`assembly_\` for assembly."

Then, you can repeat the examples and include the prefixes, if you want.

sterobin: @kalexand-rh, I agree. I addressed that in my proposed change in my comment in this PR. Does that suite what you're going for as well?

sterobin: @kalexand-rh, would you mind proposing your version of that section, similar to what I've done but how you think it should be? Always easier to have concrete proposals and then decide accordingly.

tradej: What if we changed the word _recommended_ to _suggested?_ That should clear the ambiguity IMO.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
ncbaratta: I agree with this comment.  Why not offer suggestions?

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
kalexand-rh: @tradej, s/guidelines/examples would be closer, I think. 

@sterobin, I was happy with my suggestion to the PR, but I might rewrite it like this:

```
File names:: Give the module file the same name as the anchor used in it, which is the same as or similar to the module heading. Assembly and module file names must accurately and closely reflect the title of the assembly or module.
+
While the module file name must include the anchor name, you can annotate the file names to differentiate the modules types. For example, you can a prefix with an underscore to the file name to indicate the module type in the format `prefix_file-name`. Use `con_` for concept, `ref_` for reference, `proc_` for procedure, and `assembly_` for assembly.
+
[NOTE]
====
Ensure that all members of your team use the same file naming conventions.
====
 +
 .Examples of basic file names
* `guided-decision-tables.adoc`  (Concept module)
* `creating-guided-decision-tables.adoc`  (Procedure module for creating)
* `editing-guided-decision-tables.adoc`  (Procedure module for editing)
* `guided-decision-table-examples.adoc`  (Reference module with examples)
* `guided-decision-table-columns.adoc`  (Reference module with column types)
* `guided-decision-tables.adoc`  (Assembly of guided decision table modules)
+
.Examples of annotated file names
* `con_guided-decision-tables.adoc`  (Concept module)
* `proc_creating-guided-decision-tables.adoc`  (Procedure module for creating)
* `proc_editing-guided-decision-tables.adoc`  (Procedure module for editing)
* `ref_guided-decision-table-examples.adoc`  (Reference module with examples)
* `ref_guided-decision-table-columns.adoc`  (Reference module with column types)
* `assembly_guided-decision-tables.adoc`  (Assembly of guided decision table modules)
```

sterobin: @kalexand-rh, thanks for the concrete proposal. Your "examples of basic file names" still illustrate the issue that module type annotations are there to resolve, which is that the concept file and the assembly file are identical without the annotation. Even in the basic file name scenario, annotating or otherwise distinguishing at least the assembly file is wise and often (like here) required. So we're back to my initial comment above.

mishaone: +1 to keeping the example basic as recommended. If teams want to add prefixes, they can make the choice to do so, but for the sake of standardization, it's best to be clear what the recommendation is and leaving it at that.

vikram-redhat: I agree. I don't think we need to offer examples about prefixes. If we do that, we will have another camp about why not offer suggestions for suffixes and then we will be back at PR #25. :)

Let's leave it at the basic examples. The idea here is to standardize that the file name MUST  accurately and closely reflect the headings and titles. Over and above that, teams can do what they want with the name.

tylerkelly13: @vikram-redhat your changes look good :). @tradej, @rkratky, is it correct that assembly files and modules will need to be in the same folder based on information from our tooling team? if so then a prefix on the assemblies makes sense.
I ask this because the [fedora quick docs repository structure](https://pagure.io/fedora-docs/quick-docs/blob/master/f/en-US) would make an excellent alternative to using prefixes.

rkratky: @tylerkelly13 AFAIK, there's no requirement from the tooling team regarding the placement of modules and assemblies.

I also like a structure where modules are in a subdirectory of the directory where assemblies are. But I also think this should be left to indiv. teams to decide.

mishaone: Is there a reason the historical decision was made to keep the assemblies and modules in the same folder? Modules in a separate folder and assembles in the parent folder make a lot of sense in terms of ease of finding the right sort of file (assembly vs content).

rkratky: @mishaone I don't think any decision of that sort was made -- the guide doesn't talk about directory structure. But the mod doc example suggests a subdirectory for modules: https://redhat-documentation.github.io/modular-docs/#a-simple-example-of-a-modular-guide

sterobin:
I've mentioned this twice now in this thread, but we all seem to be blowing past it. :) With this change, assembly names can potentially match existing module names. This is illustrated in the very examples introduced by this change, which cannot stand as currently updated.

Before change:
````
`con_guided-decision-tables.adoc`
`assembly_guided-decision-tables.adoc`
````

After change:
````
`guided-decision-tables.adoc`
`guided-decision-tables.adoc` 
````

I realize that this particular case would be resolved by using the exact name of the assembly title, but that would be very long ("Designing a decision service using guided decision tables") and on our team, although the anchor and file names do match exactly, they are often shortened forms of titles for several practical and long-term foresight reasons. Bear in mind that we are suggesting in this document that anchor names be "the same _or similar_" to the module headings, and that file names and anchor names match exactly. This leeway with matching headings is important. 

So we either need to think of a way to fix this problem in these examples in the doc, or prescribe a file name distinction of at least assembly files, to avoid this sort of thing.

See the first comment on this thread for details.

sterobin:
@mishaone, yes I have. The very examples used in this section are from our project, since I'm the one who originally wrote this section on anchor and file names. In my previous comment I explained why the two are the same with only the module/assembly affix differentiated. We don't need to keep this going though.

What you're suggesting is fine. Bottom line, the doc examples that result from removing affixes in this PR need to be resolved, if only just for the assembly file example, to avoid confusion. The confusion is inevitable when two different files are named the same, even if structure separates them. Our team plans to continue using assembly and module affixes (prefix for assembly, suffix for module) because of the tremendous value we've found in it, so I don't care either way, but the doc examples need to at least be usable.

tradej:
While I do recognize @sterobin's problem, I believe it shouldn't be a big one if the reader uses the now-recommended [repository structure](https://redhat-documentation.github.io/modular-docs/#a-simple-example-of-a-modular-guide) that houses assemblies separately from other modules.

sterobin:
@tradej, ack. While I agree that same-named files _can_ exist separately and that teams are welcome to do that if they wish, I don't think we should be exemplifying that in this standard-setting document.

For the sake of ending this convo, how about we just rename the assembly file in the example to `designing-guided-decision-tables.adoc` and call it good. At least that way we're not promoting same-named files (which is generally a no no), even though technically it's permitted in the file structure.

That work?

vikram-redhat:
@sterobin I never assumed that the assemblies would be in the same folder as the modules. This update only updates the suggestions around the **_naming_** of the files, not the location of these files. That is a topic for another issue :D.

As far as I know, putting everything - modules AND assemblies - in the same folder would be a terrible idea, not just because of the headache involved in managing them but also because tooling will then definitely need some sort of meta information to distinguish between what is a module and what is an assembly. But I understand that teams don't want to standardize the location of files, and so we do have to consider that possibility, even if that is a far fetched idea practically. 

So, my original suggestions still stands for this PR, but I would add the changes suggested by @kalexand-rh to reorg it a little bit so it reads better, without adding any examples for prefixes and suffixes.

tylerkelly13:
Is this issue now resolved and ready for merge?

vikram-redhat:
@tylerkelly13 - I have updated based on @kalexand-rh and @sterobin feedback.

1. Changed the wording of the note to say: Ensure that all members of your team use the same file naming conventions.

2. Changed the assembly example to: designing-guided-decision-tables.adoc.

I think we should wait a couple of days and then merge this (and then send out a blast to ccs mailing list :))

ritz303:
ACK

vikram-redhat:
@ritz303 - thanks.

I am wondering if we should also change the template file names to reflect this new guideline? The template file names start with "moduletype_" and now sort of contradict this updated guideline.

@mishaone @tylerkelly13 @tradej ?

mishaone:
+1 should change it in the templates to be consistent across the document.

sterobin:
+1 @vikram-redhat

vikram-redhat:
OK, I have added a new commit to rename file names that were being used within the guidelines themselves.

There are a few changes, so please take a look. I did do a local build and it worked fine.

I also think that we should broadcast this to CCS _before_ we merge it so people who have not been following this discussion have a chance to weigh in.

ritz303:
Okay, I've sent out an email to the CCS mailing list, giving people the opportunity to comment before we merge this PR.  I plan on merging this PR on August 10th, unless someone gives us a compelling reason not to.

laubai:
Looks good to me.

PR #62, opened 2018-07-27
Update attributes.adoc

sandrobonazzola:
Fixed a typo


ritz303:
@sandrobonazzola : Good catch!  ACK.

PR #66, opened 2018-08-03
Update anchor text

kalexand-rh:
This change updates the concept template to match the procedure and reference template. They use an underscore in front of the context reference instead of a dash.

### DIFF
--- a/modular-docs-manual/files/TEMPLATE_CONCEPT_concept-explanation.adoc
+++ b/modular-docs-manual/files/TEMPLATE_CONCEPT_concept-explanation.adoc
@@ -8,7 +8,7 @@
 // * Title: = My concept module A
 
 // The ID is used as an anchor for linking to the module. Avoid changing it after the module has been published to ensure existing links are not broken.
-[id='concept-explanation-{context}']
+[id='concept-explanation_{context}']
 // The `context` attribute enables module reuse. Every module's ID includes {context}, which ensures that the module has a unique ID even if it is reused multiple times in a guide.
 = Concept explanation
 //In the title of concept modules, include nouns or noun phrases that are used in the body text. This helps readers and search engines find the information quickly.


JStickler:
Maybe I've been hanging out with developers too long, but why not update all three to use a dash (so we don't have to hit the shift key while typing)?

ritz303:
Our docs team uses dashes, and personally I don't care which one people use.  I don't think there is any tooling requirement either way, as long as the same character is consistently used.  For consistency sake, I'm adding my ACK.  I'm targeting to merge this on August 10th.

@kalexand-rh : +1 Thanks for catching this mistake!

kalexand-rh:
I think we should take @JStickler's suggestion and make them all dashes, then - we can be consistent and make an improvement. :smile_cat:

ritz303:
@kalexand-rh : Do you want to reverse your commit, and change the rest of the templates before I accept the PR?  I'll hold off the merge until I hear back from you, and anyone else who wants to weight in.

ncbaratta:
I'm going to chime in here and say it really should be all dashes instead of us using dashes and underscores in the same value.

vikram-redhat:
@ritz303 and @kalexand-rh @JStickler @ncbaratta  - I agree we should be consistent across all templates. The Assembly and Concept templates use a dash, while the Reference and Procedure templates use an underscore.

I assume the initial reason why there is a mix of underscore and dash was to identify the separate parts in the anchor.

So:

<id>_<context>

where id would contain dashes and context would contain dashes and it would be clear to whoever is trying to figure this when it all builds that there are two components to this overall anchor and that they are joined by the underscore.

That logic makes a lot of sense to me from a purely developer/debugging purposes and I can see value there. Add parent-contexts to the whole mix and you see how we can have some value in separating the parts of the anchor.

(Maybe that wasn't the case and I am just putting my own spin on it :D. Maybe @rkratky knows more.)

With that said, I would vote to have the underscore. But if the majority disagrees, then let's use the dash. 

But let's make the change as soon as possible and make it consistent across all the templates.

sterobin:
@VladimirSlavik Your interpretation is exactly correct and was the intention I had behind it when I introduced that convention a while back. The idea is that IDs and such should be dashes all around, universally, except for the underscore division between the anchor ID and context. It makes it hugely helpful, as you'v noticed, in the rendered string: `anchor-id_context-id`, such as `guided-decision-tables-create-proc_guided-decision-tables` or whatever. Some didn't agree with having a mixed bag when I introduced this, so if we find too much push back and have to choose only one convention, I still vote dashes all around over underscores. Underscores pose problems in URL parsing and in links in some cases - some tools will break the link when an underscore is encountered, which I've seen in previous gigs. I'm surprised it's the default in Asciidoc.

vikram-redhat:
@sterobin Hi Stetson - yes, I would prefer to have the _ to delineate the multiple parts. So if there is not too much of an objection, we should stick with it.

ncbaratta:
As long as it's not required that we mix underscores and hyphens I'm fine with it.  If it's required then our upstreams do have quite a few objections (I think I raised them on another issue) because it's not a good naming practice.

jenmalloy:
Also, I think @sterobin has laid out a pretty solid argument as to why using underscores is a bad idea.

VladimirSlavik:
But the arguments about "sameness" of things in ID are going the other way that the actual practice! Having the separator of local id part and context blend together would make debugging even harder. If that should change, _it must not be the same_. If dashes are what you want there, use two of them.

vikram-redhat:
@ritz303 - Hey - I would suggest if we are going with dashes all around, then let's at least recommend that we have two dashes to separate the id and the context, as @VladimirSlavik points out. So:

`<id>--<context>`

Will really help down the road with debugging when we have 1000's of modules and assemblies.

vikram-redhat:
@VladimirSlavik - hmm. I just tested it on our docs. It works ok:
Commit: https://github.com/openshift/openshift-docs/commit/849e8ccb1e69b2358982a5bcd939c7382a95b536 and pushed live: https://access.redhat.com/documentation/en-us/openshift_container_platform/3.11/html-single/cluster_administration/#viewing-roles--and-bindings

and also using AsciiBinder (which uses AsciiDoctor) https://docs.openshift.com/container-platform/3.11/admin_guide/manage_rbac.html#viewing-roles--and-bindings

VladimirSlavik:
I dug into this some more. We're good with two dashes as long as we say no single quotes, it seems.

Two dashes work in all of these id syntaxes:
```
[[foo--bar]]
[#foo--bar]
[id="foo-bar"]
```
This one makes it an em-dash and breaks ccutil:
```
[id='foo--bar']
```
(note single quotes)

The reason I saw all these failures was that the old templates had single quotes, so most places where I'd check were the only broken case.

VladimirSlavik:
Upstream says single quotes are expected to be special and do exactly that by design: https://github.com/asciidoctor/asciidoctor/issues/2960 So yes, this should work.

vikram-redhat:
Thanks @VladimirSlavik for filing the upstream issue and following up. 

The problem now is that because the templates recommend the single quote version, most teams have been using that. If we now disallow the single quotes version, it will break docs for people who have already converted their docs to modules.

I think we may have to accept the use of single dash. :(

VladimirSlavik:
Ah, now it's my turn to not see a problem :) Changing the separator will always require a change, so if the change includes also the quotes, we're good... Or is there something else I'm not seeing?

Note that in light of the information above, it might be in the best interest of everyone to stop using the single quotes. _That_ is a change that works: possible to automate, does not break compatibility (single quotes did work so far), can be staged per any unit deemed appropriate, does not affect rendering so invisible to user ...

ritz303:
In the interest of not impacting doc teams who have already implemented the current standard, I'm just changing the "_" --> "-" in the module templates.  We want them to be consistent, as to minimize e any confusion.  This is the intended scope of this PR, PR #84 and PR #82 .  This is a recommendation and not a requirement.  I want the doc teams to have some flexibility when deciding these trivial matters.  This is basically a workaround for module reusability that the current tool chain doesn't provide.

@VladimirSlavik : If you want to continue on discussing the pros and cons of using a single quote, then please open a new issue and email the mod-docs mailing list to solicit feedback.

ritz303:
https://github.com/redhat-documentation/modular-docs/commit/18bef9b2bd89279d562cefae4cebfc02fb19c215

Issue #44, opened 2018-01-03
Modular Docs should mandate the structure of anchors


vikram-redhat:
Looking at https://redhat-documentation.github.io/modular-docs/#reusing-modules, the text recommends that 

> "When you create content in modules, you can use the same module multiple times in an assembly without having to replicate information in multiple source files. However, in order to facilitate module reuse, you must embed a document attribute variable in the anchor name for the module and then define that variable in the assembly each time the reused module appears. If the variable is not embedded and assigned, an error appears at build time reporting the duplicated anchor ID."

Shouldn't then the use of the variable be made mandatory for any new content by specifying it in the AsciiDoc Markup Conventions? https://redhat-documentation.github.io/asciidoc-markup-conventions/.

If I (working in OpenShift docs) want to include a module from the OpenStack docs, and those docs don't have the variable (context) in the headers, the books will not build due to the above mentioned issue. I would then have to request the OpenStack team to modify the headers which will add layers of issues to the whole process.

If we mandate that all new attributes have the _context_ added, this will make it easier for the future success of modular docs.

rkratky:
+1

One consideration though: wouldn't it be better then to also mandate that the `context` variable is always set in assemblies, so that when modules are included, the variable is never missing/empty?

vikram-redhat:
@rkratky - true. So all assemblies should have a `context` variable defined and all anchors should be of the form [id='section-header_{context}'] where only the `section-header` should be defined by the individual teams/authors.

ritz303:
@vikram-redhat : Yes, you make a good point.  In mandating this, we will keep things consistent and with consistency, it will be a little easier for others to build tooling around a modular content model.  I also feel this still gives the individual teams/writers some flexibility over a naming standard that makes sense for their product. 

It makes sense to modify the language in the [AsciiDoc Mark-up Conventions Guide](https://redhat-documentation.github.io/asciidoc-markup-conventions/), instead of saying:

The `{context}` variable is recommended ...

I would replace "recommended" to "required".

Also, we would need to update the NOTE under the first step in this [section of the Mod Docs Ref Guide](https://redhat-documentation.github.io/modular-docs/#reusing-modules), also changing "recommended" to "required".

I agree with rkratky, the assembly is where the `context` attribute should be defined.

+1

Best Regards,
Aron

vikram-redhat:
@ritz303 thanks. 

[...]

PR #73, opened 2018-08-20
Resolve issue #44 (b), …

tylerkelly13:
Resolve issue #44 (b), update guide and templates to indicate that context anchor needs to be redefined if a module is reused.


tylerkelly13:
I have (since) noticed that a section on this issue already exists [here](https://redhat-documentation.github.io/modular-docs/#reusing-modules). I'm leaving this pull request up because if this information is not clear as possible, it can lead to frustration and can break builds.

ritz303:
@tylerkelly13 : Since no one has made any comments,  I think we should close this PR, and if someone wants any clarification, then we can reopen it or create a new one.  Any objections to that plan?

tylerkelly13:
nope. sounds good @ritz303

Issue #71, opened 2018-08-14
parent-context doesn't not work properly in deeper structures


scholzj:
We tried to follow this documentation and use the templates for documentation in our project Strimzi. However, we have run into problems with the assembly template. The template suggests that on the beginning of every assembly, we should save the parent context:

```
:parent-context: {context}
```

And restore it at the end of every assembly.:

```
:context: {parent-context}
```

However, the template suggests that every assembly should use the same variable `parent-context`. And this works only on a very flat structure. Once you change the context on two levels you can very easily mess up the context for all the upper levels because the original context will be never restores.

As an example, you can check these two samples:
* Using an old template: [https://github.com/strimzi/strimzi-kafka-operator/tree/modular-docs-parent-context/documentation/test](https://github.com/strimzi/strimzi-kafka-operator/tree/modular-docs-parent-context/documentation/test)
* Using the latest template: [https://github.com/strimzi/strimzi-kafka-operator/tree/modular-docs-parent-context/documentation/test2](https://github.com/strimzi/strimzi-kafka-operator/tree/modular-docs-parent-context/documentation/test2)

Just check these out and open them in your editor / render them and you will see how the context is never restored properly.

So far the solution we used in our docs was:
1. Remove every `parent-context`
2. When the context is actually changed, save and restore the context, but always use the different variable to not override the other parent contexts

Can you please clarify if we are using it incorrectly or if it is a bug?

VladimirSlavik:
This would be a question primarily for @mrksu , but as his accomplice in doing this, I am sure this is indeed broken and your analysis is correct. In fact it confirms a doubt I had a few weeks earlier. (Curious that nobody ran into this, including me...)

The actual reason for doing this "parent-context" dance is that the "context" variable is global, too, and including a sub-assembly within an assembly messes context up: If the sub-assembly is followed by a module, the module gets wrong context.

An alternative approach is that in every file where there is an include::assembly followed by include::module, one must insert in between them a line setting context to the right value. It is not as neat and automatic, but works.

Sorry for brevity-hopefully it is understandable.

Asciidoc being what it is, it lacks variable scopes or stack-like structures, which leads me to believe an automated solution is not possible. We (Marek and I) agreed on this solution, believing the stacking would happen automatically. Apparently it does not.

(PS: Do we need context to agree with the assembly? It is the sole reason to have this whole context and parent-context complication.)

scholzj:
[...]

PR #83, opened 2018-09-21
Add substitution strings for the templates to be directly usable by the newdoc script

mrksu:
This PR edits template files so that they have machine-readable substitution strings in the format of the Python Template module. This makes it possible for the newdoc[1] script to download templates each time a file is generated to make sure writers always have the most up-to-date version.

Bits that have been changed:

- The module/assembly title
- The module/assembly ID
- Context-related attributes

Additionally, this PR changes the `parent-context` attribute to be assembly-specific: `parent-context-of-(assembly-ID)`. The attribute is composed automatically by newdoc when creating a new file. The reasoning is explained in [issue#71](https://github.com/redhat-documentation/modular-docs/issues/71), which is fixed with this change.

[1] https://github.com/redhat-documentation/tools/tree/master/newdoc

### DIFF
--- a/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
+++ b/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
@@ -5,7 +5,7 @@
 // Retains the context of the parent assembly if this assembly is nested within another assembly.
 // For more information about nesting assemblies, see: https://redhat-documentation.github.io/modular-docs/#nesting-assemblies
 // See also the complementary step on the last line of this file.
-ifdef::context[:parent-context: {context}]
+ifdef::context[:parent-context-of-${module_id}: {context}]
 
 // Base the file name and the ID on the assembly title. For example:
 // * file name: my-assembly-a.adoc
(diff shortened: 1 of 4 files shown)


modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
```diff
@@ -5,19 +5,19 @@
...
 // Base the file name and the ID on the assembly title. For example:
 // * file name: my-assembly-a.adoc
 // * ID: [id='my-assembly-a']
 // * Title: = My assembly A
 
 // The ID is used as an anchor for linking to the module. Avoid changing it after the module has been published to ensure existing links are not broken.
-[id='a-collection-of-modules']
-// If the assembly is reused in other assemblies in a guide, include {context} in the ID: [id='a-collection-of-modules-{context}'].
-= A collection of modules
+[id='${module_id}']
```
mrksu: I wasn't sure here: Do we want to default to assembly IDs having context or not? Alternatively, should I add an `ifdef` statement to take care of the two scenarios automatically?

mrksu:
I must admit, I feel like the substitution strings make it harder for human eyes to make sense of the templates, and they were complicated enough before I started with this.

On the other hand, if the templates are to be usable for automatic generators, which makes working with them much easier, certain parts must be unambiguous for computers, which probably means using some sort of additional markup or programming language syntax like I have used.

ritz303:
A lot has changed since this PR was submitted, so I'm closing this PR.

PR #84, opened 2018-09-21
Change - to _ for context to make it consistent in the assembly

vikram-redhat:


modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
```diff
@@ -14,7 +14,7 @@ ifdef::context[:parent-context: {context}]
 
 // The ID is used as an anchor for linking to the module. Avoid changing it after the module has been published to ensure existing links are not broken.
 [id='a-collection-of-modules']
-// If the assembly is reused in other assemblies in a guide, include {context} in the ID: [id='a-collection-of-modules-{context}'].
+// If the assembly is reused in other assemblies in a guide, include {context} in the ID: [id='a-collection-of-modules_{context}'].
```
ncbaratta: What are we making this consistent with? And why are we mixing hyphens and underscores?

ncbaratta:
See the comments on PR #66

modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
vikram-redhat: @ncbaratta all other templates use _ (underscore) for connecting the id to the {context}. I assume this was a typo when they created this template and wanted to make sure we ware consistent. If the followup from #66 is to use - (dash) then let's keep this as a - dash as well.

ritz303:
See PR #66 for more details.  Changed "_" --> "-".
https://github.com/redhat-documentation/modular-docs/commit/18bef9b2bd89279d562cefae4cebfc02fb19c215

Issue #80, opened 2018-08-30
Recommendation for a Prerequisites list that is a single prerequisite


IngridT1:
The question that came up when we were reviewing sample modular docs (flagship or topnotch modular docs) is:

* How should we format the Prerequisites section if there is a single prerequisite? Should the title still be "Prerequisites"? Should the list be bulleted even if it's just a single prerequisite? 

At this point, there's no explicit guidance. But the current recommendations from @ChristyWatkins, Betty Prioux, and me are:

* Yes, let's bullet the list even if it's just a single bullet (based on bulleted list recommendations from the IBM Style Guide).
* Let's always make the title "Prerequisites" even if there is just a single prerequisite. 

Betty, Christy, and I all prefer to keep the formatting consistent for ease of scanning. We would like this to be an official recommendation in the Mod Docs Reference Guide.

ritz303:
+1, I agree on being consistent.

mishaone:
Definitely +1 for consistency.

vikram-redhat:
+1

jenmalloy:
I'm all for consistency, but I'm not sure about this part: 
"Yes, let's bullet the list even if it's just a single bullet (based on bulleted list recommendations from the IBM Style Guide)."
The IBM Style Guide says this (p. 79 in my edition): "Use at least two list items in an ordered list. You can have one item in an unordered list if the intention is to match the format of other unordered lists in the same area.”
I don't think this case fits the second sentence though, since there are not other unordered lists "in the same area." Or maybe I am missing a key point? 
(Full disclosure: I was taught that lists must have at least two items. One item takes the form of a sentence.)

sterobin:
I agree with @jenmalloy's points, and with the notion that bulleted lists by nature need at least two items, but I'm fine going with "Prerequsities" plural with single bulleted items for the sake of consistency and simplicity, especially in cases of ifdefs like @VladimirSlavik  mentions. We can go against norm and logic so long as we do so deliberately and for a greater gain, in this case, consistency and simplicity. I think that is the case here, so +1 from me.

Just to be clear, when I say imo we'd be going against norm and logic (which I again am in favor of), here's what I mean:

[...]

PR #86, opened 2018-11-05
Update procedure module guidelines for single prereqs

bhardesty:
This is for Issue https://github.com/redhat-documentation/modular-docs/issues/80.

The consensus is that you should use a bulleted list for all prereqs, even if there's only one. This PR adds a short sentence to state this guidance explicitly.

### DIFF
--- a/modular-docs-manual/content/topics/module_guidelines-procedure.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-procedure.adoc
@@ -24,6 +24,8 @@ Prerequisites are conditions that must be satisfied before the user starts the p
 
 Focus on relevant prerequisites that users might not otherwise be aware of. Do not list obvious prerequisites.
 
+Use a bulleted list for prerequisites, even if you only have a single prerequisite.
+
 [discrete]
 == Procedure Body
 The procedure consists of one or more steps required to complete the procedure. Each step describes one action.


sterobin:
@bhardesty, @ritz303, @IngridT1, understood. Did we also decide to use plural "Prerequisties" even when just one? That and the bullets issue were the two parts to this question, so just confirming.

IngridT1:
@sterobin Yes, always use the plural, "Prerequisites", even when you have just one.

bhardesty:
@sterobin @IngridT1 Another option would be to use a phrase that works for both singular and plural prereqs (like, "Before you begin"). But considering that many of us have already implemented the "Prerequisites" term, I think it would be pretty disruptive to change it.

sterobin:
Good point, @bhardesty. To me, though, "Before you begin" and "Prerequisites" aren't quite the same thing, at least in my cases. IMO, "Before you begin" conveys active steps the user should do pertaining to the upcoming procedure but that really aren't part of that procedure (cases are rare, but I've encountered them), while "Prerequisites" are check-off things that should be done prior to the procedure. I don't have a great example on hand (I'd need to dig a bit) and what I'm saying might not make any sense, but I guess it's just to say that there's another reason to not replace "Prerequisites".

ritz303:
@bhardesty : Looks good, and thanks for taking care of this! ACK.  I plan on merging this into master next Monday (11/12).

PR #90, opened 2019-01-31
Issue 53: add FCC template-compliant examples and associated code examples

jenmalloy:
Not sure I've chosen the best format to show both the built example and the associated code, so please comment on that with suggestions for improvement.


kalexand-rh:
Thanks @jenmalloy! Merging.

PR #92, opened 2019-03-11
standardizing -{context}

### LINKED ISSUE #58: Suggestions for template improvements: mostly ID-related
The information about IDs in the templates can be confusing, at least for AsciiDoc novices. 

Here are a few suggestions for improvements:

- The [id='my-concept-module-a'] does not match the [id='concept-explanation-{context}']. I think it would be more effective for both examples to match...right now the syntax is too different.
- The education provides a best practice/recommendation of a prefix for file names (con_, etc), and the template says the ID should match. Therefore, the best practice should be reflected in the template. The prefix should be used for the filename and ID. (Even if you note that it is best practice, follow what your library does, etc.)
- I think it would be better to get rid of the "my-xyz..." as an example and using some more real examples, such as "what-is-foo" or "starting-the-foo" or "foo-parameters." Using a name that is more real for each module type is a simple reminder of what that module's focus should be and how the module title should be worded.
- For someone who has not been working in AsciiDoc yet, I wasn't sure how to interpret the {context} right away--so it took a few moments to process. That's another place where a "real" ID example could help the writers less familiar with AsciiDoc realize that the {context} is code that stays (not some kind of variable to replace): [id='con_what-is-foo-{context}']. For newbies, it would also probably be better to explain it a bit more: "The context attribute (coded as "-{context}" at the end of the ID string) is included in every ID...." or something similar.

kalexand-rh:
Fixes #58 

All of the sample files now use `-{context}` in the anchor IDs, so we should update the rest of the topics.

I've updated this PR to include fixes for the remaining issues in #58.

### DIFF
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -3,7 +3,7 @@
 
 To optimize modular documentation, follow these guidelines for naming module anchors and files:
 
-Anchor names:: Provide an anchor in the format `+++[id='anchor-name-{context}']+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. `+++{context}+++` is a variable whose value you define in the assembly. Give the anchor the same or similar name as the module heading. Separate the words in the anchor with dashes:
+Anchor names:: Provide an anchor in the format `+++[id='anchor-name-{context}']+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. `+++{context}+++` is a variable whose value you define in the assembly. Give the anchor the same or similar name as the module heading. Separate the words in the anchor with hyphens:
 +
 --
 [source]
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -3,12 +3,12 @@
 
 To optimize modular documentation, follow these guidelines for naming module anchors and files:
 
-Anchor names:: Provide an anchor in the format `+++[id='anchor-name_{context}']+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. Give the anchor the same or similar name as the module heading, separated by dashes:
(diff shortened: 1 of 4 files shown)


modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -3,7 +3,7 @@
 
 To optimize modular documentation, follow these guidelines for naming module anchors and files:
 
-Anchor names:: Provide an anchor in the format `+++[id='anchor-name-{context}']+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. Give the anchor the same or similar name as the module heading, separated by dashes:
+Anchor names:: Provide an anchor in the format `+++[id='anchor-name-{context}']+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. `+++{context}+++` is a variable whose value you define in the assembly. Give the anchor the same or similar name as the module heading. Separate the words in the anchor with dashes:
```
jenmalloy: This is a tiny nit, but "dashes" should be "hyphens" throughout

jenmalloy:
@kalexand-rh Thanks so much for making these edits!

kalexand-rh:
Happy to help, @jenmalloy. I replaced "dashes" with "hyphens" and an happy to fix any other nits that you see.

modular-docs-manual/content/topics/module_reusing-modules-procedure.adoc
```diff
@@ -17,7 +17,7 @@ This error is resolved by adding and defining a document variable.
 [discrete]
 .Procedure
 
-. In the module file that will be reused, add the `+++{context}+++` suffix with an underscore to the anchor name in the format `[id='anchor-name_+++{context}'+++]`.
+. In the module file that will be reused, add the `+++{context}+++` suffix with a dash to the anchor name in the format `[id='anchor-name-+++{context}'+++]`.
```
adahms: One instance of 'dash' instead of 'hyphen'.

kalexand-rh: Fixed! Thank you.

adahms:
Thanks @kalexand-rh!

Looks good to me; merging.

PR #93, opened 2019-03-12
you can't use xrefs in modules

### LINKED ISSUE #64: Cross-referencing between modules should not be done within modules
Hi everyone,

I'm currently in the process of putting together a user story that is predominantly made up of modules from a number of user stories. The biggest issue I'm having by far is reusing modules that link to other modules.

When a cross-reference (xref) exists that points to another module, I have two options:

1) I can use conditionals to remove that link from my particular user story.
2) I can pull in the module that it links to.

The second option tends to create a cascade effect of pulling in more and more content that might only be vaguely related to the user story I'm trying to create.

The first option can result in spaghetti code and possibly break other stories, particularly where assemblies are included within other assemblies. It also doesn't really scale when we think about reuse across products and all possible user stories.

The current modular docs [templates](https://github.com/redhat-documentation/modular-docs/blob/master/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc), as well as the [Modular Documentation Reference Guide](https://redhat-documentation.github.io/modular-docs/#creating-procedure-modules), advise including links to additional resources closely related to the contents of the module ("other documentation resources (such as assemblies or modules)..") within the module.

I am proposing that instead of providing cross-references to other modules within modules, **they should only be added to the assembly file(main.adoc).** Doing this results in the link only appearing in the assembly where the linked module already exists and allows more flexible module reuse across multiple assemblies without needing to add conditionals everywhere.

For example:

**include::product-migration-guide/migration-configure-kie-server-proc.adoc[leveloffset=+1]**
For information about installing {PRODUCT} on {EAP_LONG}, see xref:eap-dm-install-pro[].

renders the following sentence as though it were a part of the preceding module:
"For information about installing Red Hat Process Automation Manager on Red Hat JBoss Enterprise Application Platform, see Installing Business Central from the ZIP file "

**TL;DR - We should only be cross-referencing between modules in the actual assembly files, rather than in modules themselves to make module reuse more flexible.**

kalexand-rh:
Fixes #64

### DIFF
--- a/modular-docs-manual/content/topics/module_guidelines-concept.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-concept.adoc
@@ -37,5 +37,7 @@ See also link:http://www.informationmapping.com/fspro2013-tutorial/infotypes/inf
 == Concept Additional Resources
 
 The optional additional resources list links to other material closely related to the contents of the concept module, for example, other documentation resources.
+Because of current tooling limitations, you cannot include xref links to other content in your collection in the concept module.
+You can include xref links in the assembly that contains the concept.
 
 Focus on relevant resources that are likely to be of immediate interest to the user. Do not list resources for completeness.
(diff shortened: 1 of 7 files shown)


adahms:
Thanks @kalexand-rh!

Looks good to me.

adahms:
Now merging these changes.

PR #121, opened 2020-06-30
Resolves Issues 109, 111, and 115

emmurphy1:
[Rendered HTML](http://file.ork.redhat.com/~emmurphy/file-names-and-anchors#module_anchor-and-file-names-concept)

[Modular Documentation Review Board Notes](https://docs.google.com/document/d/10uGTmEPeFgYvOZQZnfo-VGrRltRSbAwU70eppULahas/edit#heading=h.kat2ctqr49lg)

This update reflects recent decisions by the Modular Documentation Review Board:
File names must be prefixed with proc-, con-, or ref-, or proc_, con_, or ref_, or each module must include a module type variable.
While not a requirement of Pantheon, it is good practice to include the prefix assembly- or assembly_ with assembly module files.

After consultation with the Pantheon engineering team, they agreed that we can change :pantheon-module-type: to :system-module-type: to make it more generic.

Anchor names should match files names without the .adoc extension, and with the addition of _{context}.

Please review these changes and post your comments or approval.

### DIFF
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -27,7 +27,7 @@ Create assembly and module file names that accurately and closely reflect the ti
 Do not include special characters in file names. Ensure that all members of your team use the same file naming conventions.
 ====
 
-These file naming guidelines are optional but highly recommended. However, if your team does not include the module prefixes in file names followed by either a hyphen (-) or an underscore (_), you must include one of the following variables in each concept, procedure, and reference module:
+These file naming guidelines are optional but highly recommended. However, if your team does not include the module prefixes in file names followed by either a hyphen (-) or an underscore (_), include one of the following variables in each concept, procedure, and reference module:
 
 [source]
 ----
@@ -35,8 +35,6 @@ These file naming guidelines are optional but highly recommended. However, if yo
 :system-module-type: PROCEDURE
 :system-module-type: REFERENCE
 ----
-
-NOTE: This is a requirement if you will publish on Pantheon V2. For more information, see xref:introduction_{context}[].
 
 .Anchors
 At the top of every module, provide an anchor in the format `+++[id="filename_{context}"]+++` where `filename` is the exact name of the file, without the file extension (`.adoc`). Module anchors are necessary so that Asciidoctor can identify the module when the module is reused or cross-referenced. The  `context` variable is defined in each assembly module, for example `:context: my-context-value`. When you build an assembly, the value of the `context` variable replaces `{context}` in each module anchor ID and appears in generated URL.
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -62,7 +62,7 @@ The guided decision tables feature works similarly to ...
 .Example 2. Procedure Module
 [source]
 ----
-[id="pro-creating-guided-decision-tables_{context}"]
+[id="proc-creating-guided-decision-tables_{context}"]
 = Creating Guided Decision Tables
 
 You can use guided decision tables to ...
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -4,7 +4,7 @@
 To optimize modular documentation, follow these guidelines for naming files and creating anchors:
 
 .File names
-Assembly and module file names should accurately and closely reflect the title of the assembly or module. File names should have the format `prefix-filename.adoc` or `prefix_filename.adoc` where `prefix` is one of the following module prefixes:
+Create assembly and module file names that accurately and closely reflect the title of the assembly or module. Create file names with the format `prefix-filename.adoc` or `prefix_filename.adoc` where `prefix` is one of the following module prefixes:
 
 * `con`: Concept module prefix
 * `proc`: Procedure module prefix
@@ -13,11 +13,13 @@ Assembly and module file names should accurately and closely reflect the title o
 
 .Examples
 * `con-guided-decision-tables.adoc`  (Concept module)
+* `con_guided-decision-tables.adoc`  (Concept module)
 * `proc-creating-guided-decision-tables.adoc`  (Procedure module for creating)
-* `proc-editing-guided-decision-tables.adoc`  (Procedure module for editing)
+* `proc_creating-guided-decision-tables.adoc`  (Procedure module for creating)
 * `ref-guided-decision-table-examples.adoc`  (Reference module with examples)
-* `ref-guided-decision-table-columns.adoc`  (Reference module with column types)
+* `ref_guided-decision-table-examples.adoc`  (Reference module with examples)
 * `assembly-designing-guided-decision-tables.adoc`  (Assembly of guided decision table modules)
+* `assembly_designing-guided-decision-tables.adoc`  (Assembly of guided decision table modules)
 
 
 [NOTE]
@@ -26,8 +28,6 @@ Do not include special characters in file names. Ensure that all members of your
 ====
 
 These file naming guidelines are optional but highly recommended. However, if your team does not include the module prefixes in file names followed by either a hyphen (-) or an underscore (_), you must include one of the following variables in each concept, procedure, and reference module:
-
-NOTE: This is a Pantheon 2 requirement.
 
 [source]
 ----
@@ -39,8 +39,8 @@ NOTE: This is a Pantheon 2 requirement.
 
 
 .Anchors
-At the top of every module, provide an anchor in the format `+++[id="filename_{context}"]+++` where `filename` is the exact name of the file, without the file extension (`.adoc`). Module anchors are necessary so that the module can be identified by Asciidoctor when reused or cross-referenced. The `{context}` variable is intentionally not defined in the module anchor ID. You define a context value in each assembly, in the format `:context: my-context-value`. This enables the context variable in each module to be populated according to the relevant assembly.
---
+At the top of every module, provide an anchor in the format `+++[id="filename_{context}"]+++` where `filename` is the exact name of the file, without the file extension (`.adoc`). Module anchors are necessary so that Asciidoctor can identify the module when the module is reused or cross-referenced. The  `context` variable is defined in each assembly module, for example `:context: my-context-value`. When you build an assembly, the value of the `context` variable replaces `{context}` in each module anchor ID and appears in generated URL.
+
 [source]
 ----
 [id="filename_{context}"]
@@ -53,16 +53,34 @@ The first sentence of the topic.
 [source]
 ----
 [id="con-guided-decision-tables_{context}"]
-
 = Guided Decision Tables
 
 The guided decision tables feature works similarly to ...
 ----
 
-.Example 2. Procedure Module
+.Example 2. Concept Module
+[source]
+----
+[id="con_guided-decision-tables_{context}"]
+= Guided Decision Tables
+
+The guided decision tables feature works similarly to ...
+----
+
+
+.Example 3. Procedure Module
 [source]
 ----
 [id="proc-creating-guided-decision-tables_{context}"]
+= Creating Guided Decision Tables
+
+You can use guided decision tables to ...
+----
+
+.Example 4. Procedure Module
+[source]
+----
+[id="proc_creating-guided-decision-tables_{context}"]
 = Creating Guided Decision Tables
 
 You can use guided decision tables to ...
@@ -74,7 +92,7 @@ You can use guided decision tables to ...
 The format defined here is recommended because it is the most stable and versatile of anchor formats, and supports variables that enable topics to be reused and cross-referenced properly. For details, see xref:reusing-modules[]. Other anchor formats include `+++[[anchor-name]]+++` and `+++[#anchor-name]+++`, but these formats either do not support variables for content reuse or do not support certain character types, such as periods. These limitations cause errors at build time.
 ====
 
---
+
 
 .Additional Resources
 
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -36,7 +36,7 @@ These file naming guidelines are optional but highly recommended. However, if yo
 :system-module-type: REFERENCE
 ----
 
-
+NOTE: This is a requirement if you will publish on Pantheon V2. For more information, see xref:introduction_{context}[].
 
 .Anchors
 At the top of every module, provide an anchor in the format `+++[id="filename_{context}"]+++` where `filename` is the exact name of the file, without the file extension (`.adoc`). Module anchors are necessary so that Asciidoctor can identify the module when the module is reused or cross-referenced. The  `context` variable is defined in each assembly module, for example `:context: my-context-value`. When you build an assembly, the value of the `context` variable replaces `{context}` in each module anchor ID and appears in generated URL.
--- a/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
+++ b/modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
@@ -1,14 +1,49 @@
-[id="anchor-and-file-names"]
-= Anchor Names and File Names
+[id="module_anchor-and-file-names-concept"]
+= File Names and Anchors
 
-To optimize modular documentation, follow these guidelines for naming module anchors and files:
+To optimize modular documentation, follow these guidelines for naming files and creating anchors:
 
-Anchor names:: Provide an anchor in the format `+++[id="anchor-name_{context}"]+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. `+++{context}+++` is a variable whose value you define in the assembly. Give the anchor the same or similar name as the module heading. Separate the words in the anchor with hyphens and use an underscore to separate the `+++{context}+++` part:
-+
+.File names
+Assembly and module file names should accurately and closely reflect the title of the assembly or module. File names should have the format `prefix-filename.adoc` or `prefix_filename.adoc` where `prefix` is one of the following module prefixes:
+
+* `con`: Concept module prefix
+* `proc`: Procedure module prefix
+* `ref`: Reference module prefix
+* `assembly`: Assembly module prefix
+
+.Examples
+* `con-guided-decision-tables.adoc`  (Concept module)
+* `proc-creating-guided-decision-tables.adoc`  (Procedure module for creating)
+* `proc-editing-guided-decision-tables.adoc`  (Procedure module for editing)
+* `ref-guided-decision-table-examples.adoc`  (Reference module with examples)
+* `ref-guided-decision-table-columns.adoc`  (Reference module with column types)
+* `assembly-designing-guided-decision-tables.adoc`  (Assembly of guided decision table modules)
+
+
+[NOTE]
+====
+Do not include special characters in file names. Ensure that all members of your team use the same file naming conventions.
+====
+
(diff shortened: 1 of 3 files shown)


sterobin:
@emmurphy1 Would you mind also updating the concrete anchor ID examples in the [reusing assemblies section](http://file.ork.redhat.com/~emmurphy/file-names-and-anchors#reusing-modules), just for consistency? I wrote those in parallel with the former file name guidelines, so would be ideal to update them to stay aligned.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -1,14 +1,49 @@
...
+* `ref-guided-decision-table-columns.adoc`  (Reference module with column types)
+* `assembly-designing-guided-decision-tables.adoc`  (Assembly of guided decision table modules)
+
+
+[NOTE]
+====
+Do not include special characters in file names. Ensure that all members of your team use the same file naming conventions.
+====
+
+These file naming guidelines are optional but highly recommended. However, if your team does not include the module prefixes in file names followed by either a hyphen (-) or an underscore (_), you must include one of the following variables in each concept, procedure, and reference module:
```
sterobin, line 28: Maybe we want to make it a little stronger than optional. How about this:

>As an alternative to these file-naming conventions, if your team cannot include the module prefixes in file names followed by either a hyphen (-) or an underscore (_), you can include one of the following variables in each concept, procedure, and reference module:

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -1,14 +1,49 @@
...
+
+NOTE: This is a Pantheon 2 requirement.
```
sterobin, line 30: I think maybe nix this note? It's just irrelevant in Pantheon 1, plus all of this is really in preparation for Pantheon 2. If you want to leave it, I suggest just rewording to "This is supported in Pantheon 2 only." instead of calling it a requirement. Makes it seem at a glance that even if you do name your files with the correct prefix, still required to have these variables. IDK. Your call.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -1,14 +1,49 @@
...
+[id="module_anchor-and-file-names-concept"]
+= File Names and Anchors
 
-To optimize modular documentation, follow these guidelines for naming module anchors and files:
+To optimize modular documentation, follow these guidelines for naming files and creating anchors:
 
-Anchor names:: Provide an anchor in the format `+++[id="anchor-name_{context}"]+++` for every module so that it can be identified by Asciidoctor when reused or cross-referenced. `+++{context}+++` is a variable whose value you define in the assembly. Give the anchor the same or similar name as the module heading. Separate the words in the anchor with hyphens and use an underscore to separate the `+++{context}+++` part:
-+
+.File names
+Assembly and module file names should accurately and closely reflect the title of the assembly or module. File names should have the format `prefix-filename.adoc` or `prefix_filename.adoc` where `prefix` is one of the following module prefixes:
```
vikram-redhat, line 7: Why or? The prefix should be joined to the filename using the underscore, right?

prefix_<filename>.adoc

Further, we should remove 'should' from here as this is only a recommendation, and not a hard requirement as teams can choose not to do this and go the attribute route.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -1,14 +1,49 @@
-[id="anchor-and-file-names"]
-= Anchor Names and File Names
+[id="module_anchor-and-file-names-concept"]
```
vikram-redhat, line 1: Should this be con_anchor-and-file-names and the file also renamed to con_anchor-and-file-names.adoc to keep it consistent with the guideline that it is prescribing?

Further, we should add a guideline that terms like 'and', 'the' should not be put into filenames. Otherwise with this recommendation of keeping filenames the same as titles, we will end up with really long filenames like: https://github.com/openshift/openshift-docs/blob/master/modules/developer-cli-odo-creating-and-deploying-a-nodejs-application-with-odo.adoc

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -1,14 +1,49 @@
...
+* `ref`: Reference module prefix
+* `assembly`: Assembly module prefix
+
+.Examples
+* `con-guided-decision-tables.adoc`  (Concept module)
+* `proc-creating-guided-decision-tables.adoc`  (Procedure module for creating)
+* `proc-editing-guided-decision-tables.adoc`  (Procedure module for editing)
+* `ref-guided-decision-table-examples.adoc`  (Reference module with examples)
+* `ref-guided-decision-table-columns.adoc`  (Reference module with column types)
+* `assembly-designing-guided-decision-tables.adoc`  (Assembly of guided decision table modules)
```
vikram-redhat, line 20: I highly encourage this to be of the form: prefix_filename.adoc. We have had endless discussions on this in this forum already and the final decision was to separate parts of a filename with underscores.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
vikram-redhat: @sterobin Sorry Stetson. I think this should be diluted and both options presented as an equal candidate. If tooling is not forcing our hand on this, then both options are equal.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
vikram-redhat: Again, this goes hand in hand with the above suggestion. This is a hard Pantheon 2 requirement, so we should mention it, but we should present both options as equal.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -1,14 +1,49 @@
...
+----
+:system-module-type: CONCEPT
+:system-module-type: PROCEDURE
+:system-module-type: REFERENCE
+----
+
+
+
+.Anchors
+At the top of every module, provide an anchor in the format `+++[id="filename_{context}"]+++` where `filename` is the exact name of the file, without the file extension (`.adoc`). Module anchors are necessary so that the module can be identified by Asciidoctor when reused or cross-referenced. The `{context}` variable is intentionally not defined in the module anchor ID. You define a context value in each assembly, in the format `:context: my-context-value`. This enables the context variable in each module to be populated according to the relevant assembly.
```
vikram-redhat, line 42: Why does it have to be filename?

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: This section is about file naming guidelines and anchor names. We have noted the requirement to specify the module type if a group chooses a different file naming convention. I suggest we leave the wording as it is.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: Both hyphens and underscores are valid prefix separators. 'Should' implies that these are recommendations.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: Yes, this entire repository needs to be updated to reflect these recommendations and this will be done is a separate PR. 
We do not recommend that file or anchor names match the module headings exactly. Some groups will choose to do that, however.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: This is not a hard requirement.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: @vikram-redhat Understood, thanks.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: @vikram-redhat Sure, I was saying more to modify so that it's clearer that the prefix or variable is a requirement, one or the other. This not here to me makes it seem like the variable is the hard requirement, regardless of the prefix (meaning, even if you use the prefix, also have to use the variable as a requirement, which isn't what we mean). I suggest we just explain that requirement as part of the instruction in the text, or if we like the note, then put it somewhere earlier so it's clear it applies to both options.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: @vikram-redhat Just the majority preference, and has a lot of conceptual and practical merit or many. One example is asciidoc processing tooling (rendering, validating, splitting, etc) is benefitted by this practice. This is definitely no reason to enforce it, but just like most FCC guidelines, mainly for those who just need to know a way to go and don't have strong reasons otherwise.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
kalexand-rh: I disagree that "should" implies that it's optional. It implies that it's required. I suggest following the IBM Style Guide's guidance here:

> ! should v
Always prefer a more direct verb or way of saying something. Omit “should,” or use “must” or “might,” as
appropriate.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
kalexand-rh: @sterobin, I think that we need a separate section to discuss the Pantheon 2 requirements. I think it would be clearer to present the file name guidance as a best practice here in isolation. In the requirements section, we could explain that you must mark the module file types. One way is to use the recommended naming practices. If you can't use the recommended naming practices, then you need to add these attributes.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
oraNod: +1 to underscore as delimiter

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
oraNod, line 42: s/top/start 

"where `filename` matches the name of the file without the extension (`.adoc`)" 

Lots of passive voice here and I find it difficult to read. 

"Module anchors are necessary so that the module can be identified by Asciidoctor when reused or cross-referenced. The `{context}` variable is intentionally not defined in the module anchor ID. You define a context value in each assembly, in the format `:context: my-context-value`. This enables the context variable in each module to be populated according to the relevant assembly."

"Anchors uniquely identify modules for re-use and cross-reference. In modules, `{context}` is a variable that you define in an assembly. When you then build the documentation, those variable definitions replace `{context}` in the module anchors."

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
sterobin: @kalexand-rh Sure, I'll leave it up to @emmurphy1 if she wants to do that or just mention in passing as she has.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
vikram-redhat: Ah @sterobin that is fine and understandable. I am very well aware of the benefits this gives, as we use the same guideline in OpenShift docs (https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#module-file-metadata) and have built many tooling practices around that for docs.openshift.com.

This is the right thing to do for the benefits you mention. I think I have combined my concerns about long filenames which I mentioned in another part  and that then leading to really long module ids in this section. If we add/modify guideline to make the filenames not needing to be exactly as the title, I think we are good to ignore this part.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
rkratky: @vikram-redhat, I think we should not call out "Pantheon" at all in these guidelines. They are available and promoted for use in upstream docs projects and in other organizations. References to internal-only tooling and/or conditioning any of the rules/guidelines on the use of a particular version of internal tooling turns this content into a RH-only thing. I don't think we want that.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
rkratky: +1 While not a hard requirement, let's at least use the underscore in the presented example.

sterobin: Sounds funny to say, but maybe we just list examples for both ways, the list with hyphens and the list with underscores :) Since we are in fact suggesting both are fine, and we seem unable to come to a consensus of which examples to show here, perhaps just show both and then everyone's happy. Just a thought.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
rkratky: @vikram-redhat, I don't agree that words like 'and' or 'the' should not be in file names. Long file names are not good practice, I agree, but so are not long headings in general. Writers should -- wherever possible -- avoid long headings (which in turn eliminates long file names). Omitting 'and' or 'the' does not help that much in this regard.

More importantly, we have tooling (such as `newdoc` or `nebel`) for automatic generation of module and assembly files (pre-filled templates). These tools automate much of the tedious manual work a writer would need to do when starting on new docs. While it would be possible to modify the tools to omit a defined list of words from file names, it seems like an unnecessary exercise.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
vikram-redhat: I agree with @rkratky and @kalexand-rh and combining those two suggestions would make sense. Have a separate Pantheon 2 section which essentially lists the requirements for making these guidelines work with RH tooling.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
vikram-redhat: @rkratky Ok and that is fine. So the guidance (or a hard requirement?) is that the file names/anchor names/module names must match exactly, excluding the "proc|con|ref" bit?

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
theashiot: +1 @kalexand-rh 

Also, since the sub-section about anchors addresses the reader directly ("At the top of every module, provide an anchor in the.."), perhaps this section can also be written that way - doing so will also avoid the use of "should".

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: OK, that makes sense. I'll remove the Pantheon 2 note, @vikram-redhat . @kalexand-rh  would you like to create the Pantheon 2 section? However, before you do, about @rkratky's point to  not include anything about Pantheon 2 in this guide.  @rkratky would you feel comfortable if we  add a separate Pantheon section or do you feel strongly that we shouldn't mention Pantheon at all?

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: @rkratky  @vikram-redhat  the guidance is that it is preferable (not required) for the anchor to match the file name exactly, including the prefix, and that the file name/anchor should approximate the actual module title.  However, I believe RHEL uses a tool that creates module IDs and file names from the module title so they include the small words and they follow a process of ensuring that the module title and anchor match exactly even when the module title changes. That is fine, but it's not what the board agreed to recommend. There are other variations on file naming that don't break Pantheon and unless these variations cause a Pantheon or other problem (ultimate single repository?) groups are free to deviate from the recommended formats as long as they include the prefix or the variable in each module.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: Good suggestion! I've rewritten and removed all 'shoulds' from the section.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: OK, I've added two sets of examples.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
emmurphy1: @oraNod I've rewritten.

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
```diff
@@ -1,14 +1,49 @@
...
-[id="anchor-name_{context}"]
+:system-module-type: CONCEPT
+:system-module-type: PROCEDURE
+:system-module-type: REFERENCE
+----
+
+
+
+.Anchors
+At the top of every module, provide an anchor in the format `+++[id="filename_{context}"]+++` where `filename` is the exact name of the file, without the file extension (`.adoc`). Module anchors are necessary so that Asciidoctor can identify the module when the module is reused or cross-referenced. The  `context` variable is defined in each assembly module, for example `:context: my-context-value`. When you build an assembly, the value of the `context` variable replaces `{context}` in each module anchor ID and appears in generated URL.
```
oraNod, line 42:
```suggestion
At the top of every module, provide an anchor in the format `+++[id="filename_{context}"]+++` where `filename` is the exact name of the file, without the file extension (`.adoc`). Module anchors are necessary so that Asciidoctor can identify the module when it is reused or cross-referenced. The  `context` variable is defined in each assembly module, for example `:context: my-context-value`. When you build an assembly, the value of the `context` variable replaces `{context}` in each module anchor ID and appears in the generated URL.
```

modular-docs-manual/content/topics/module_anchor-and-file-names-concept.adoc
vikram-redhat: @ emmurphy1 right so this goes back to @rkratky's point about this being a RH specific guideline then if we are only including it for accomodating Pantheon and the RHEL tool.

emmurphy1:
@vikram-redhat @rkratky @oraNod @theashiot @adahms I have rewritten the introduction to include a description of Pantheon.

emmurphy1:
@redhat-documentation/fcc-review-board if you haven't done so already, please either approve or add comments to this PR. I need four more approvers to merge. Note that I have rewritten the Introduction section to include Pantheon.

emmurphy1:
@redhat-documentation/fcc-review-board I removed reference to Pantheon from this content as discussed at the last review board meeting. Pantheon-specific information will go in the Pantheon 2 User Guide. I need three more approvals to merge. Would you please review and either comment or approve if you haven't approved already? Thanks.

rolfedh:
lgtm

rolfedh (approved):
lgtm

rolfedh (approved):
lgtm

oraNod (approved):
LGTM. Thanks @emmurphy1

Preeticp:
LGTM

Preeticp (approved):
LGTM

emmurphy1:
@adahms  this issue is now resolved. Please record it in the FCC Decision Register.

PR #134, opened 2020-09-30
Recommendation to use second level heading syntax for assembly-level prerequisites

### LINKED ISSUE #124: Issues with the assembly-level prerequisites heading
This issue originated in https://github.com/redhat-documentation/modular-docs/issues/97 but deserves its own separate issue.

After discussing this with @kalexand-rh, I have a better understanding of the item @ncbaratta mentioned. Although it deserves fuller discussion, here is the issue in summary:

Currently, `.Prerequisite` appears in both TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc and TEMPLATE_PROCEDURE_doing-one-procedure.adoc.

The problem is that, in TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc, `.Prerequisite` is inconsistent with `== Additional resources (or Next steps)`.

Changing `.Prerequisites` to `== Prerequisites` would do the following:

- Eliminate a style inconsistency.
- Make it clear that the prerequisites apply to all of the modules, not just the first one. (With `.Prerequisites`, it's unclear what the scope of the prerequisites is. One possible interpretation is that it applies only to the first module instead of all of them.)
- Enable "Prerequisites" for assemblies near the top of the content hierarchy to show up in the TOC.

Issue https://github.com/redhat-documentation/modular-docs/issues/97 also discusses arguments against making this change. I thought it would be important to take a more complete look at the proposed change so we could have a more nuanced discussion of its benefits and drawbacks. I'll be glad to describe the issue in more detail at an upcoming meeting. 

For an example of what the current assembly-level `.Prerequisites` looks like in our documentation, please see: https://docs.openshift.com/container-platform/4.5/installing/installing_aws/installing-aws-default.html

Preeticp:
This issue fixes #124 as per discussions with the team and as per comments on the issue.

### DIFF
--- a/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
@@ -16,7 +16,9 @@ Consider rewording the user story to write the assembly introduction, for exampl
 [discrete]
 == Assembly Prerequisites
 
-Prerequisites are conditions that must be satisfied before the user can start following the assembly.
+Prerequisites are conditions that must be satisfied before the user can start following the assembly and are applicable to all the modules in the assembly.
+
+Use the second level heading syntax for the Prerequisites section in the assembly so that it is displayed in the table of contents and is consistent with the Additional resources or Next steps sections in the assembly.
 
 // [bhardest] - We have a lot of xref-ing in these guidelines. A better approach might be to create a "snippets" .adoc file with snippets of common content (for example, the content about writing prerequisites, which applies to multiple sections). Then we can just include the relevant content from the snippets file wherever it's needed.
 // [asteflova] - Let's do this after we finish reviewing the guidelines for procedures and assemblies.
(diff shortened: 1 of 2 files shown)


Preeticp:
@rkratky @sterobin @emmurphy1 @theashiot can you all please review this PR. Thank you!

rolfedh (approved):

sterobin (approved):
@Preeticp Good with me for the reasons we've discussed on the call, though I got the sense that we still don't have a majority agreement yet. Will see if we can get enough approvals here. Thanks for this!

oraNod (approved):
@Preeticp good suggestion IMO

theashiot (approved):

rkratky (approved):

kalexand-rh (commented):
LGTM!

naciao (approved):

mrksu:
I've updated the assembly template in `newdoc` to match this change. Commit: https://github.com/redhat-documentation/newdoc/commit/9e271823a556de55d7d7ac3df789dc10081a800c 

The updated template is available in `newdoc` version 2.6.1, which I'm now releasing.

Issue #135, opened 2020-09-30
Change format of Additional Resources/Next Steps to bulleted links only in the templates


emmurphy1:
For the Additional Resources or Next Steps sections in the assembly, concept, and procedure modules, currently some writers provide lead-in text and some writers provide just links:

Additional resources

* For information about BLAH A, see <link>.
* For information about BLAH B, see <link>.

Additional resources

* <link>
* <link>

NOTE: In PV2, anything under Additional Resources or Next Steps (tagged with the new 'addtional-resources' tag) will appear in a box under the module headed Related Content. However the PDF will have the heading in the source file (Additional resources or Next steps).

rkratky:
@emmurphy1, I believe this should be left to the writer to determine what's best for their particular case. In other words, let's only add this as a non-binding recommendation.

While I agree that oftentimes, links alone suffice (and are easier to parse when quickly scanning the content), there are cases where some context is called for. For example, when adding a link to a resource whose connection to the matter at hand is not immediately obvious (i.e. not apparent from the title of the linked content), it's useful to include a note about.

So, if the module in question is called **Configuring XYZ**, then it makes no sense to say:

* _For more information about XYZ configuration, see **Overview of XYZ configuration options**_

Here, the prefixed "context" is redundant.

However, when linking to something else, it's a different story:

* _To understand how XYZ configuration affects performance, review **Debugging Go applications**_

Here, the context makes sense, IMO.

emmurphy1:
Conversation carried over from PR 130:

rolfedh 13 days ago  • 
 
"X is installed. For information about installing X, see " contains redundant information and surplus words :-). We could simplify this by converting the prerequisite into a link. The link text should state the prerequisite, not the title of the topic.

Suggested change 
* X is installed. For information about installing X, see <link>.
* link:<url>[X is installed.] 
 
 @rolfedh rolfedh 13 days ago  • 
 
The proposed format is also more consistent with the new "Additional resources"

 
 @emmurphy1 emmurphy1 13 days ago  Author  Member
OK, Well, I just thought I was giving a practical example. The focus of this PR is to get the attributes in so I've reinstated the original prerequisite guidance and we can look at this later if we want.

 
 @stoobie stoobie 2 days ago  Member
[...]

PR #136, opened 2020-09-30
On Hold: issue-135 Change format of Additional Resources/Next Steps to bulleted links only in the templates

emmurphy1:
[Issue 135](https://github.com/redhat-documentation/modular-docs/issues/135)


sterobin (approved):
@emmurphy1 Agreed, and approved. I've found this to be best anyway personally, and if anytime you need to give additional context to the link you're providing, often it is an indication that you might need to provide that link directly in the content. But then that too is often discouraged from mod docs perspective, but even so, the vast majority of the cases you can still get away with just a link in the additional resources and it's still understandable and consistent.

rolfedh (approved):

ncbaratta (approved):
LGTM

emmurphy1:
On hold pending the outcome of this ticket:
https://projects.engineering.redhat.com/browse/CCS-4121

emmurphy1:
The review board decided that this is a style issue and should be left to individual writers. In addition, lists of links without context can be problematic from an accessibility POV. I am closing this PR without merging.

Issue #137, opened 2020-09-30
Clarify use of sub-headings in concepts and references


sterobin:
Subheadings in con and ref modules comes up a lot in team discussions and in my mod doc repos course. So we should add a brief paragraph to each module type in the ref guide noting that concepts and references can have sub-headings if needed for organizing the content better and if the sub-headings are truly helpers and cannot/should not be separate modules. This is not true for procedures, which should always contain the one heading plus labels.

Also clarify that when sub-headings are used, the understood default to date is to `[discrete]` the heading,  if you have a need to display the sub-headings in the TOC, permitted to not use `[discrete]`. But the use case must be deliberate, such as API endpoints that can be extensive. You don't always want them as individual ref modules, but also want them displayed in TOC for easier nav. So okay to have a parent ref module that contains all endpoints as subsections without `[float]`.

sterobin:
Submitted PR #140

sterobin:
Resolved in above-mentioned PR.

rkratky:
@sterobin, apologies for being so late to the party.

I believe we should resolve this issue by providing solid info on what constitutes a module. I think we already have that in the mod docs glossary:

_*Module*_

_An independent, self-contained chunk of information with a well-organized structure. Each module is information that a reader can understand and use by itself. A writer can use a module as a standalone article or as part of a larger body of work (an "Assembly"). A module does not depend on other documents, but it can refer the reader to other documents as additional resources. Because a module is self-contained, it must have a clear title that briefly and clearly summarizes the subject of the module. Moreover, because modules are written as context-free elements independent of other modules, they are re-usable. One module can be part of multiple assemblies. [...]_

This is very clear. Perhaps we need to highlight this in another part of the guide, or include it more prominently in training sessions. We should not try to invent rules or overly restrict the way writers use their judgement. As long as writers follow the cited guideline, their modules will conform to mod docs requirements -- they will be reusable and self-contained. Whether the writer chooses to use subtitles or not is up to them and their particular content situation.

sterobin:
[...]

PR #140, opened 2020-10-14
Add guidance about subheadings in cons and refs.

sterobin:
Addresses issue #137

### DIFF
--- a/modular-docs-manual/content/topics/module_guidelines-concept.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-concept.adoc
@@ -33,6 +33,37 @@ Do not include any instructions to perform an action, such as executing a comman
 Action items belong in procedure modules.
 See also link:http://www.informationmapping.com/fspro2013-tutorial/infotypes/infotype2.html[The Six Information Types] at _informationmapping.com_ for ways to present different types of conceptual information: principle, concept, structure, process, fact.
 
+If the concept module is large and complex, either split the concept module into multiple standalone concept modules or use subheadings in the module to structure the content for improved reader navigation. If you use subheadings in a concept module, you can use the `[discrete]` tag to exclude the heading from the table of contents, if needed. However, in many cases, subheadings are helpful to include in the table of contents.
+
+The following examples illustrate a discrete and a standard subheading:
+
+.Example discrete subheading excluded from table of contents
+[source]
+----
+= My concept module
+
+Concept introduction and body
+
+[discrete]
+== My concept module subheading
+
+More concept body content
+----
+
+.Example subheading included in table of contents
+[source]
+----
+= My concept module
+
+Concept introduction and body
+
+== My concept module subheading
+
+More concept body content
+----
+
+NOTE: You can use subheadings in concept or reference modules, but not in procedure modules.
+
 [discrete]
 == Concept Additional Resources
 
--- a/modular-docs-manual/content/topics/module_guidelines-concept.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-concept.adoc
@@ -33,7 +33,7 @@ Do not include any instructions to perform an action, such as executing a comman
 Action items belong in procedure modules.
 See also link:http://www.informationmapping.com/fspro2013-tutorial/infotypes/infotype2.html[The Six Information Types] at _informationmapping.com_ for ways to present different types of conceptual information: principle, concept, structure, process, fact.
 
-If the concept module is large and complex, either split the concept module into multiple standalone concept modules or use subheadings in the module to structure the content for improved reader navigation. If you use subheadings in a concept module, you can use the `[discrete]` tag to exclude the heading from the table of contents, if needed. However, in many cases, subheadings are helpful to include in the table of contents.
+If the concept module is large and complex, consider splitting the concept module into multiple standalone concept modules. If you cannot split the module into meaningful standalone modules, consider using subheadings in the module to structure the content for improved reader navigation. If you use subheadings in a concept module, you can add the `[discrete]` tag to exclude each subheading from the table of contents, if needed. However, in many cases, subheadings are helpful to include in the table of contents to improve content searchability.
 
 The following examples illustrate a discrete and a standard subheading:
 
--- a/modular-docs-manual/content/topics/module_guidelines-reference.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-reference.adoc
@@ -20,3 +20,34 @@ To make the reference data easier to scan, organize it in a logical order (such 
 * link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#tables[Tables]
 
 If you have a large volume of the same type of information to document, use a structure into which the information details can fit, and then document each logical unit of information as one reference module. For example, think of man pages, which document very different information details, but which still use consistent titles and formats to present those details in a uniform information structure.
+
+If the reference module is large and complex, either split the reference module into multiple standalone reference modules or use subheadings in the module to structure the content for improved reader navigation. If you use subheadings in a reference module, you can use the `[discrete]` tag to exclude the heading from the table of contents, if needed. However, in many cases, subheadings are helpful to include in the table of contents.
+
+The following examples illustrate a discrete and a standard subheading:
+
+.Example discrete subheading excluded from table of contents
+[source]
+----
+= My reference module
+
+Reference introduction and body
+
+[discrete]
+== My reference module subheading
+
+More reference body content
+----
+
+.Example subheading included in table of contents
+[source]
+----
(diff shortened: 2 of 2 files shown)


emmurphy1 (approved):

oraNod (changes requested):
I feel like the guidance here should be to split modules that are "large and complex" where units of information can be standalone, rather than combining pieces of information into mega-modules that use sub-headings. What I mean is that we should be encouraging writers to break modules into standalone chunks. 

I also think the "discrete" attribute should be used where the information constitutes a section that is directly and uniquely related to the module but not necessarily standalone. The "discrete" attribute is like a "float" heading. It's a division within a module that helps to organize a subset of information.

Hopefully what I'm saying makes sense. It just seems like this PR is missing some nuance and not being prescriptive enough.

fbolton (approved):
LGTM!

sterobin:
@oraNod Thanks for the feedback. I guess your points are exactly what this PR is intentionally trying to make less prescriptive. What I mean is, there is a tradition in CCS that cons and refs really shouldn't have subheadings and that writers should favor splitting modules into standalone units, just as you're saying. As a result, many writers are creating artificial modules that really aren't standalone just for the sake of avoiding subheadings and to comply with the preference to break modules apart. 

Regarding the discrete/float heading, your preference to retain that tag is again what this PR is attempting to loosen. Meaning, writers are likewise under the impression that if you do use subheadings in cons and refs, they should be discrete. The problem there is that there are many cases where content should not be discrete so that it appears in the TOC, but where that subheading isn't a candidate for it's own module. A recurring example is API reference content, especially endpoints, which in some cases make more sense to be collected as part of the same reference module, but would benefit from being listed in the TOC (non-discrete).

This is all highly subjective and relies on the writers' judgement, so I suppose my purpose here is to open things up a bit more to writers' needs and not be quite so prescriptive. But if you and others feel that your perspective is still the commonly shared feeling of what we ought to be conveying, then basically I can just disregard this PR and let that tradition continue on as it is. We'll see what others think and go from there.

oraNod:
Thanks for explaining the purpose a bit more @sterobin Maybe I'm going in the total opposite direction but I still think the wording makes things a little open to interpretation.

modular-docs-manual/content/topics/module_guidelines-concept.adoc
```diff
@@ -33,6 +33,37 @@ Do not include any instructions to perform an action, such as executing a comman
 Action items belong in procedure modules.
 See also link:http://www.informationmapping.com/fspro2013-tutorial/infotypes/infotype2.html[The Six Information Types] at _informationmapping.com_ for ways to present different types of conceptual information: principle, concept, structure, process, fact.
 
+If the concept module is large and complex, either split the concept module into multiple standalone concept modules or use subheadings in the module to structure the content for improved reader navigation. If you use subheadings in a concept module, you can use the `[discrete]` tag to exclude the heading from the table of contents, if needed. However, in many cases, subheadings are helpful to include in the table of contents.
```
oraNod, line 36: @sterobin I'm thinking about something along these lines (not this wording, I'm just throwing my thoughts out here):

If you are writing a concept module that has a lot of content, you might decide to break that content into separate modules. This approach is good if modules can standalone, meaningful units of information. Creating separate modules lets you easily move content around and re-use it. It's also easier to update, edit, and delete content that is broken into separate modules. 

However, as an alternative to creating separate modules, you can add subheadings to nest content that is directly, and uniquely, related to the concept you're writing about. Subheadings are a good way to organize large concepts into sections that improve reader navigation. You can also add the `[discrete]` attribute to subheadings so that they do not appear in the table of content. However, doing this might reduce retrieve-ability as the headings do not display in the high-level TOC so users can't find floating sections when scanning the document.

sterobin: Thanks @oraNod , yeah makes sense. I'll see how I can tweak, a bit later.

theashiot:
I guess the modules in this PR are the perfect example of why it's good to have subheadings. The modules "Concept Module Guidelines" and "Reference Module Guidelines" without the subheadings would become difficult to read. Splitting the modules into, for example, "Reference Introduction", "Reference Body" would not make sense.  

The changes look good to me!

theashiot (approved):

modular-docs-manual/content/topics/module_guidelines-concept.adoc
sterobin: @oraNod I reworded a bit to lean more toward modularizing, or as an alternative, subdividing. I've intentionally kept it fairly simple though. I don't know that we want to get into more detail or be any more forceful than than about splitting into modules. If anything, the point of this PR is to encourage writers to consider subheadings more often so that they'll stop over-splitting concept modules into artificial mini-modules, which is what I'm observing. So I've proposed the rewording as a compromise. Let me know what you think.

sterobin:
@theashiot Exactly! +1 Thanks for the review.

modular-docs-manual/content/topics/module_guidelines-concept.adoc
```diff
@@ -33,6 +33,37 @@ Do not include any instructions to perform an action, such as executing a comman
 Action items belong in procedure modules.
 See also link:http://www.informationmapping.com/fspro2013-tutorial/infotypes/infotype2.html[The Six Information Types] at _informationmapping.com_ for ways to present different types of conceptual information: principle, concept, structure, process, fact.
 
+If the concept module is large and complex, consider splitting the concept module into multiple standalone concept modules. If you cannot split the module into meaningful standalone modules, consider using subheadings in the module to structure the content for improved reader navigation. If you use subheadings in a concept module, you can add the `[discrete]` tag to exclude each subheading from the table of contents, if needed. However, in many cases, subheadings are helpful to include in the table of contents to improve content searchability.
```
oraNod, line 36:
```suggestion
If the concept module is large and complex, consider splitting the concept module into multiple standalone concept modules. If you cannot split the module into meaningful standalone modules, consider using subheadings in the module to structure the content for improved reader navigation. If you use subheadings in a concept module, you can add the `[discrete]` tag to render subheadings as sections that are not included in the table of contents. However you should consider if this negatively affects content searchability. In most cases you want to include subheadings in the table of contents so readers can more easily scan.
```

sterobin: @oraNod In terms of minimalism, I feel like that proposed rewording is a wordier way of saying the same thing. Is there something specific that's missing that you think I should be adding? I don't see anything new in the rewording, just an expanded, less concise version of the same ideas. If I'm crazy, please let me know :)

oraNod: ok, disregard @sterobin

oraNod (approved):
:+1:

Preeticp (approved):
The recommendations make sense to me.

ncbaratta (approved):
LGTM

modular-docs-manual/content/topics/module_guidelines-reference.adoc
```diff
@@ -20,3 +20,34 @@ To make the reference data easier to scan, organize it in a logical order (such
...
+= My reference module
+
+Reference introduction and body
+
+== My reference module subheading
+
+More reference body content
+----
+
+NOTE: You can use subheadings in concept or reference modules, but not in procedure modules.
```
rkratky, line 53: Why not?

rkratky:
@sterobin, please, see my comment in the orig. issue: https://github.com/redhat-documentation/modular-docs/issues/137

modular-docs-manual/content/topics/module_guidelines-reference.adoc
sterobin: This was explicitly requested when mod docs was first kicked off by Christy and Sheldon, where subheadings were fine for cons and refs but not for procs. The rationale was/is, given the nature of procs and their need to be centered on a clear, focused task, if you need to divide a proc into different subheadings, then it should almost always (or simply always) be separate procedures. Cons and refs don't have the same stringency because concepts and references can require a lot of content but still be a part of the same overall idea that might not be wise to split up into sub-modules. That's highly subjective of course, as we discussed with Don in this thread, and there is surely an edge case for a multi-headed procedure, but to date, the recommendation from the beginning has been to permit subheads in cons and refs but not procs. So what my PR does is simply clarity that, not add anything new or change anything that wasn't already the case from the beginning.

Now, if we're having second thoughts, we should discuss as a team in that case. If this simply opens a can of worms, I can always revert this PR altogether. But this comes up repeatedly with those I work with and what writers are doing to address this need is overly divide content into artificial and overly minute con and ref modules simply in order to avoid using subheadings, which isn't necessary.

PR #159, opened 2021-03-15
Adding _module-type metadata flag to template files

laubai:
The current Pantheon v2 migration checklist lists the _module-type flag as a requirement, but this did not yet appear in the template. Placing in what I think is the appropriate location in the template files; please edit as required.


emmurphy1:
@laubai there are two ways for Pantheon 2 to identify the module type: by file name or by variable. I have asked Levi to clarify this option in the PV2 MS3 checklist. We discussed this at the review board and decided not to add the variable to the templates because it is an option. The choice is documented in the reference guide:
https://redhat-documentation.github.io/modular-docs/#module_anchor-and-file-names-concept

However, we agreed to add a comment to the templates that includes this option and I will tag you on that PR.

emmurphy1:
@laubai here is the new PR:
https://github.com/redhat-documentation/modular-docs/pull/164

Issue #160, opened 2021-03-17
Clarify rules for text following [role="_abstract"] in the templates


stoobie:
Currently the templates say this:

`Write a short introductory paragraph that provides an overview of the module. Procedure modules should include  the steps that users perform and address user motivation.The text that immediately follows the `[role="_abstract"]` tag is used for search metadata.`
Please clarify what delineates an abstract paragraph. 
- Does the abstract need to be on a single line or can it span multiple lines?
- Can a procedure abstract include bullets?
- Also note need to add a space between "motivation." and "The"

emmurphy1:
https://github.com/redhat-documentation/modular-docs/pull/163

emmurphy1:
@stoobie , the board decided to keep PV2-specific information out of the templates as much as possible so we have provided minimal instructions about what should go in the short introductions. When you include this  in the PV2 User Guide, please ask users to include key words that can be useful in a search. This is the main reason for the abstract tag as I understand it. We can work with you to help shape the content if you like. @adahms, I have merged the PR and am now closing this issue.

Issue #155, opened 2021-03-03
Inconsistent module IDs in the procedure module template


yetanothertw:
In the procedure module template [1], in the "Base the file name and the ID on the module title. For example:" section (block comment), the recommendation for the module ID format is (line 8):
* ID: [id="doing-procedure-a_{context}"]  

and then after this section (block comment) the example uses the following ID (line 14):
[id="proc-doing-one-procedure_{context}"]

I think they should both be [id="proc-doing-one-procedure_{context}"], so line 8 needs to be updated.

[1] https://raw.githubusercontent.com/redhat-documentation/modular-docs/master/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc

yetanothertw:
@emmurphy1 @adahms @sterobin just making sure you're aware of this issue, that is if you can validate that it is an issue indeed.
Many thanks

adahms:
Thanks for raising this issue, @yetanothertw!

A review of the module suggests the examples are consistent with themselves - i.e. the context and title in the top example match one another, and the context and title in the actual module piece match one another - but are not consistent with one another.

@emmurphy1 @sterobin - would making these two sets the same reduce some confusion?

emmurphy1:
https://github.com/redhat-documentation/modular-docs/pull/163

yetanothertw:
Thanks, Emily!

emmurphy1:
@adahms this issue was updated in PR 164, and merged.

PR #163, opened 2021-03-30
WIP: issues 160 and 155 updated the template ID and abstract instructions

emmurphy1:
This PR includes [issue 160](https://github.com/redhat-documentation/modular-docs/issues/160) and [issue 155](https://github.com/redhat-documentation/modular-docs/issues/155).

### DIFF
--- a/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
+++ b/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
@@ -28,7 +28,7 @@ The `context` attribute enables module reuse. Every module ID includes {context}
 ////
 
 [role="_abstract"]
-This paragraph is the assembly introduction. It explains what the user will accomplish by working through the modules in the assembly and sets the context for the user story the assembly is based on. The text that immediately follows the `[role="_abstract"]` tag is used for search metadata.
+This paragraph is the assembly introduction. It explains what the user will accomplish by working through the modules in the assembly and sets the context for the user story the assembly is based on. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
 
 == Prerequisites
 
--- a/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
+++ b/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
@@ -28,7 +28,7 @@ The `context` attribute enables module reuse. Every module ID includes {context}
 ////
 
 [role="_abstract"]
-This paragraph is the assembly introduction. It explains what the user will accomplish by working through the modules in the assembly and sets the context for the user story the assembly is based on. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
+This paragraph is the assembly introduction. It explains what the user will accomplish by working through the modules in the assembly and sets the context for the user story the assembly is based on.
 
 == Prerequisites
 
--- a/modular-docs-manual/files/TEMPLATE_CONCEPT_concept-explanation.adoc
+++ b/modular-docs-manual/files/TEMPLATE_CONCEPT_concept-explanation.adoc
@@ -21,7 +21,7 @@ In the title of concept modules, include nouns or noun phrases that are used in 
 ////
 
 [role="_abstract"]
-Write a short introductory paragraph that provides an overview of the module. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
+Write a short introductory paragraph that provides an overview of the module. 
 
 The contents of a concept module give the user descriptions and explanations needed to understand and use a product.
 
--- a/modular-docs-manual/files/TEMPLATE_CONCEPT_concept-explanation.adoc
+++ b/modular-docs-manual/files/TEMPLATE_CONCEPT_concept-explanation.adoc
@@ -21,7 +21,7 @@ In the title of concept modules, include nouns or noun phrases that are used in 
 ////
 
 [role="_abstract"]
-Write a short introductory paragraph that provides an overview of the module. The text that immediately follows the `[role="_abstract"]` tag is used for search metadata.
+Write a short introductory paragraph that provides an overview of the module. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
 
 The contents of a concept module give the user descriptions and explanations needed to understand and use a product.
 
--- a/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
+++ b/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
@@ -1,25 +1,25 @@
 ////
 Base the file name and the ID on the module title. For example:
 * file name: proc-doing-procedure-a.adoc
-* ID: [id="doing-procedure-a_{context}"]
+* ID: [id="proc-doing-procedure-a_{context}"]
 * Title: = Doing procedure A
 
 The ID is an anchor that links to the module. Avoid changing it after the module has been published to ensure existing links are not broken.
 ////
 
-[id="proc-doing-one-procedure_{context}"]
+[id="proc-doing-procedure-a_{context}"]
 
 ////
 The `context` attribute enables module reuse. Every module ID includes {context}, which ensures that the module has a unique ID even if it is reused multiple times in a guide.
 ////
 
-= Doing one procedure
+= Doing procedure A
 ////
 Start the title of a procedure module with a verb, such as Creating or Create. See also _Wording of headings_ in _The IBM Style Guide_.
 ////
 
 [role="_abstract"]
-Write a short introductory paragraph that provides an overview of the module. Procedure modules should include  the steps that users perform and address user motivation.The text that immediately follows the `[role="_abstract"]` tag is used for search metadata.
+Write a short introductory paragraph that provides an overview of the module. Procedure modules should include  the steps that users perform and address user motivation. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
 
 .Prerequisites
 
--- a/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
+++ b/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
@@ -19,7 +19,9 @@ Start the title of a procedure module with a verb, such as Creating or Create. S
 ////
 
 [role="_abstract"]
-Write a short introductory paragraph that provides an overview of the module. Procedure modules should include  the steps that users perform and address user motivation. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
+Write a short introductory paragraph that provides an overview of the module.
+
+Procedure modules should include the steps that users perform and address user motivation.
 
 .Prerequisites
 
--- a/modular-docs-manual/files/TEMPLATE_REFERENCE_reference-material.adoc
+++ b/modular-docs-manual/files/TEMPLATE_REFERENCE_reference-material.adoc
@@ -7,7 +7,7 @@ Base the file name and the ID on the module title. For example:
 The ID is an anchor that links to the module. Avoid changing it after the module has been published to ensure existing links are not broken.
 ////
 
-[id="ref-reference-material_{context}"]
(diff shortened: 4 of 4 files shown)


ncbaratta (approved):
lgtm

stoobie (approved):
Lgtm

sterobin (approved):

rolfedh (approved):

fbolton (approved):
LGTM

modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
```diff
@@ -28,7 +28,7 @@ The `context` attribute enables module reuse. Every module ID includes {context}
 ////
 
 [role="_abstract"]
-This paragraph is the assembly introduction. It explains what the user will accomplish by working through the modules in the assembly and sets the context for the user story the assembly is based on. The text that immediately follows the `[role="_abstract"]` tag is used for search metadata.
+This paragraph is the assembly introduction. It explains what the user will accomplish by working through the modules in the assembly and sets the context for the user story the assembly is based on. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
```
rkratky, line 31: I don't agree with this as it is purely implementation-specific. There's no such distinction about paragraphs in AsciiDoc or Asciidoctor.

```suggestion
This paragraph is the assembly introduction. It explains what the user will accomplish by working through the modules in the assembly and sets the context for the user story the assembly is based on.
```

modular-docs-manual/files/TEMPLATE_CONCEPT_concept-explanation.adoc
```diff
@@ -21,7 +21,7 @@ In the title of concept modules, include nouns or noun phrases that are used in
 ////
 
 [role="_abstract"]
-Write a short introductory paragraph that provides an overview of the module. The text that immediately follows the `[role="_abstract"]` tag is used for search metadata.
+Write a short introductory paragraph that provides an overview of the module. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
```
rkratky, line 24:
```suggestion
Write a short introductory paragraph that provides an overview of the module.
```

modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
```diff
@@ -1,25 +1,25 @@
 ////
 Base the file name and the ID on the module title. For example:
 * file name: proc-doing-procedure-a.adoc
-* ID: [id="doing-procedure-a_{context}"]
+* ID: [id="proc-doing-procedure-a_{context}"]
```
rkratky, line 4: Why would this be included in the ID. Unlike the file name or the "module type" attribute, the ID is actually surfaced to the end-user through anchors. And the user has no interest in internal code-specific elements, such as the module type. Consider the following two URL examples:

`https://www.example.com/docs/prod-doing-stuff/`

and:

`https://www.example.com/docs/doing-stuff/`

Which one makes more sense?


```suggestion
* ID: [id="doing-procedure-a_{context}"]
```

modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
```diff
@@ -1,25 +1,25 @@
...
 
-= Doing one procedure
+= Doing procedure A
 ////
 Start the title of a procedure module with a verb, such as Creating or Create. See also _Wording of headings_ in _The IBM Style Guide_.
 ////
 
 [role="_abstract"]
-Write a short introductory paragraph that provides an overview of the module. Procedure modules should include  the steps that users perform and address user motivation.The text that immediately follows the `[role="_abstract"]` tag is used for search metadata.
+Write a short introductory paragraph that provides an overview of the module. Procedure modules should include  the steps that users perform and address user motivation. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata.
```
rkratky, line 22:
```suggestion
Write a short introductory paragraph that provides an overview of the module. Procedure modules should include  the steps that users perform and address user motivation.
```

modular-docs-manual/files/TEMPLATE_REFERENCE_reference-material.adoc
```diff
@@ -7,7 +7,7 @@ Base the file name and the ID on the module title. For example:
 The ID is an anchor that links to the module. Avoid changing it after the module has been published to ensure existing links are not broken.
 ////
 
-[id="ref-reference-material_{context}"]
+[id="ref-my-reference-a_{context}"]
```
rkratky, line 10:
```suggestion
[id="my-reference-a_{context}"]
```

modular-docs-manual/files/TEMPLATE_REFERENCE_reference-material.adoc
```diff
@@ -17,7 +17,7 @@ In the title of a reference module, include nouns that are used in the body text
 ////
 
 [role="_abstract"]
-Write a short introductory paragraph that provides an overview of the module. The text that immediately follows the `[role="_abstract"]` tag is used for search metadata. A reference module provides data that users might want to look up, but do not need to remember.
+Write a short introductory paragraph that provides an overview of the module. The text that immediately follows the `[role="_abstract"]` tag up to a line break is used for search metadata. A reference module provides data that users might want to look up, but do not need to remember.
```
rkratky, line 20:
```suggestion
Write a short introductory paragraph that provides an overview of the module. A reference module provides data that users might want to look up, but do not need to remember.
```

rkratky (changes requested):

rolfedh (changes requested):
I support the changes that Robert is proposing, which remove some tooling or implementation-specific details. The revised suggestions help preserve the usability of this content for writers.
For the `id` parameter, perhaps we should simply say that it should match the file name.

stoobie:
> I support the changes that Robert is proposing, which remove some tooling or implementation-specific details. The revised suggestions help preserve the usability of this content for writers.
> For the `id` parameter, perhaps we should simply say that it should match the file name.

@rolfedh See issue 160. When I was using the template, it was not clear to me what technically comprises an abstract, and so I did in incorrectly. Where am I (or anyone) supposed to find this info?

Also, is the [role="_abstract"] tag _not_ implementation-specific? If it is not, then what are the vanilla rules for this tag?

If you're not going to include some guidance here, then you need to provide it somewhere. How is anyone supposed to know how to do it?

mikemckiernan (approved):
/lgtm properly.

modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
emmurphy1: @rkratky this is a convention that the group agreed on before you join.

modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
rkratky: > Then the solution is to indicate that this is implementation-specific, not to not provide guidance.

I don't think we should go that way. There are potentially many implementations. The standard/guidelines should be implementation-agnostic. Listing all implementation-specific details or considerations is not scalable and would only introduce clutter.

> See issue 160.

It's valid question, but I don't think there should be any hard limitation -- this should be up to the writer.

> These templates are part of FCC, which is a combination of several specific implementations.

I believe this is at the heart of this misunderstanding.

The templates are not part of FCC. The templates are a part of 'modular documentation' guidelines. FCC is a project at Red Hat that makes heavy use of modular docs principles. But these guidelines are an upstream project that can be used by anyone, regardless of what company they work for or what publication toolchain they use. Therefore, the guidelines (and templates) should not be modelled after FCC and/or Pantheon needs.

To provide an analogy: it would be like wanting to modify the DITA guidelines because a Red Hat tool asked for it.

> These templates include other items that are not "vanilla Asciidoc", including the :context: attribute.

If there are things included that are not part of vanilla AsciiDoc, then it should be fixed. However, that is not the case with the `:context:` attribute. Attributes and their use are a standard part of AsciiDoc. There's nothing non-standard in that, and all existing implementations of the AsciiDoc language support this mark-up element and translate it without problems.

modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
rkratky: Respectfully, it's a convention that must have been added *after* the templates were first created. I was there :)

But thanks for the explanation. I will open a separate issue.

modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc
rolfedh: The changes @rkratky proposes still provide guidance for writing an abstract. 

@stoobie What do you mean by "this"?  If you mean `[role="_abstract"]`, it is a PV2-specific requirement. Keeping it in the mod docs templates "harms" writers who must go back and remove it to comply with their teams' requirements to *not include PV2-specific elements.* 

For example, the OpenShift Docs team has proscribed `[role="_abstract"]` to avoid its incremental inclusion in their doc set. Writers like myself who accidentally include it in their content must go back and remove it afterwards. The team's intention is to avoid "messes" that must be cleaned up when they run scripts to update their whole doc set for PV2. I expect other doc teams are also adopting this approach of avoiding incremental PV2 changes so they can efficiently migrate their content when it is time to do so.

I'll check to see if there is a definitive source for PV2 requirements like this one.

rolfedh: This PV2 migration checklist covers elements that are required: https://docs.engineering.redhat.com/pages/viewpage.action?spaceKey=FCCADOPT&title=Pv2+MS3+migration+checklists

rolfedh:
> > I support the changes that Robert is proposing, which remove some tooling or implementation-specific details. The revised suggestions help preserve the usability of this content for writers.
> > For the `id` parameter, perhaps we should simply say that it should match the file name.
> 
> @rolfedh See issue 160. When I was using the template, it was not clear to me what technically comprises an abstract, and so I did in incorrectly. Where am I (or anyone) supposed to find this info?
> 
> Also, is the [role="_abstract"] tag _not_ implementation-specific? If it is not, then what are the vanilla rules for this tag?
> 
> If you're not going to include some guidance here, then you need to provide it somewhere. How is anyone supposed to know how to do it?

@stoobie, please see my response to these same questions above: https://github.com/redhat-documentation/modular-docs/pull/163#discussion_r631018633

emmurphy1:
@stoobie, @rkratky makes a very good point. With his guidance, we are trying to make the templates as independent as possible from Pantheon. Management asked us to include the abstract tags in the templates because they are essential  to PV2 and do not interfere with other uses of the templates AFAIK. I agree that we need to give CCS folks more information about what PV2 expect under those tags. Perhaps the best place for that information is the PV2 Users Guide? WDYT. Note that I have updated the templates with Robert's suggestions.

mikemckiernan:
I'll hazard my opinion.  I've come around to thinking that the PV2 docs are the best place to cover the reqs that PV2 drives to source files.

For example, I find this req to specify the abstract very tedious and a poor use of human time.  I proposed (and think it's in use) that PV2 [just use the first para](https://github.com/redhataccess/pantheon/pull/607) when the abstract attribute is missing--or intentionally omitted because there are more valuable things to do.

So, @stoobie, folks tuned into PV2 might know about a change like that one and might be in a position to update the PV2 docs to describe the PV2-specific req, but the mod-docs folks might not be in the information line-of-sight.

That's my opinion.  It and a quarter will get you a phone call.

emmurphy1:
@mikemckiernan  AFAIK, the abstract tag is required by PV2. Although that might change in the future by one mechanism or another, right now it has to be there.  Management requested that we include the tags in the templates because to not include them but instruct writers to add them would create confusion. I believe that @rkratky acquiesced to this compromise because the tag does not hinder non-Pantheon 2 users. I agree that if PV2 could identify the first para as the abstract without the tag, removing the tag would be ideal. Let's revisit this issue if the requirement changes.

yzimmerm:
lgtm

yzimmerm (approved):

stoobie:
@emmurphy1 @mikemckiernan @rolfedh This is, in fact, something that we can mention in the Pv2 docs, and this is actually a good time for this.

Issue #112, opened 2020-02-06
nested vs leveloffset 


pwright:
I have somehow got the impression that we *must* nest assemblies..
I'm reporting the following as a bug:

There's no mention of leveloffset in the reference guide, but  lots of discussion of nested assemblies.

sterobin:
Discussed with @pwright in chat and addressed with PR #166

pwright:
node https://raw.githubusercontent.com/redhat-documentation/modular-docs/master/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc

emmurphy1:
@adahms I've merged PR 166 which address this issue.

PR #166, opened 2021-05-12
Add line about leveloffset.

sterobin:
Addresses #112 . Apparently there has historically been a misunderstanding among CCS teammates that leveloffset is prohibited in mod docs. @pwright as part of his submitted issue suggested the simple added text shown in this PR to clarify that. The issue is already addressed in the assembly template itself here: https://raw.githubusercontent.com/redhat-documentation/modular-docs/master/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc, so this small text addition should remove all doubt and close the loop. 

I realize that we're trying to avoid getting into ascii syntax suggestions etc. as part of mod docs, but Paul made a strong case that there is confusion out there. So if this clarifies that, then seems to deserve the mention.

### DIFF
--- a/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
@@ -27,7 +27,7 @@ Use the second level heading syntax for the Prerequisites section in the assembl
 [discrete]
 == Assembly Modules
 
-List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly.
+List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. A `leveloffset` can provide a method of introducing hierarchy, such as `module-file.adoc[leveloffset=+2]`.
 
 [discrete]
 == Assembly Additional Resources
--- a/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
@@ -27,7 +27,16 @@ Use the second level heading syntax for the Prerequisites section in the assembl
 [discrete]
 == Assembly Modules
 
-List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute to set the hierarchy of the module relative to the assembly. In most cases, create modules whose title is an H1 and include those modules by using the format `module-file.adoc[leveloffset=+1]`. Otherwise, if necessary, you can use other `leveloffset` values, such as `module-file.adoc[leveloffset=+2]`, to modify the structure of your module content.
+To include modules in an assembly, use the Asciidoc  link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include directive]. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. Use the `leveloffset` attribute to set the hierarchy of the module relative to the assembly, as shown in the following example:
+
+.Level offset for module files
+----
+file1.adoc[leveloffset=+1]
+file2.adoc[leveloffset=+2]
+file3.adoc[leveloffset=+3]
+----
+
+NOTE: All module and assembly titles must use the H1 heading designation (`=`).
 
 [discrete]
 == Assembly Additional Resources
--- a/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
@@ -27,7 +27,7 @@ Use the second level heading syntax for the Prerequisites section in the assembl
 [discrete]
 == Assembly Modules
 
-List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. A `leveloffset` can provide a method of introducing hierarchy, such as `module-file.adoc[leveloffset=+2]`.
+List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute, such as `module-file.adoc[leveloffset=+1]` or `module-file.adoc[leveloffset=+2]`, to set the hierarchy of the module relative to the assembly.
 
 [discrete]
 == Assembly Additional Resources
--- a/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
@@ -36,7 +36,7 @@ file2.adoc[leveloffset=+2]
 file3.adoc[leveloffset=+3]
 ----
 
-NOTE: All module and assembly titles must use the H1 heading designation (`=`).
+All module and assembly titles must use the H1 heading designation, such as `= My heading`.
 
 [discrete]
 == Assembly Additional Resources
--- a/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
+++ b/modular-docs-manual/content/topics/module_guidelines-assembly.adoc
@@ -27,7 +27,7 @@ Use the second level heading syntax for the Prerequisites section in the assembl
 [discrete]
 == Assembly Modules
 
-List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute, such as `module-file.adoc[leveloffset=+1]` or `module-file.adoc[leveloffset=+2]`, to set the hierarchy of the module relative to the assembly.
+List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute to set the hierarchy of the module relative to the assembly. In most cases, create modules whose title is an H1 and include those modules by using the format `module-file.adoc[leveloffset=+1]`. Otherwise, if necessary, you can use other `leveloffset` values, such as `module-file.adoc[leveloffset=+2]`, to modify the structure of your module content.
 
 [discrete]
 == Assembly Additional Resources


pwright (approved):

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
```diff
@@ -27,7 +27,7 @@ Use the second level heading syntax for the Prerequisites section in the assembl
 [discrete]
 == Assembly Modules
 
-List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly.
+List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. A `leveloffset` can provide a method of introducing hierarchy, such as `module-file.adoc[leveloffset=+2]`.
```
rolfedh, line 30:
```suggestion
List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute, such as `module-file.adoc[leveloffset=+1]`,  to set the hierarchy of the module relative to the assembly.
```
Reworded this to speak more directly to the writer's needs.

sterobin: @rolfedh Okay on bullet one. On bullet two, that's kind of the point of this note, to specify that leveloffset of 2 is fine to use when needed. There are many cases when a module is a subchapter after another module, such as a high-level concept module that has relevant proc modules that are related to it. There is a misnomer in the past and maybe still out there that leveloffset of 2 or 3 is forbidden, which isn't true.

rolfedh (approved):
/lgtm

rkratky:
@emmurphy1 @pwright Thanks for the addition. I'd say we can be even more authoritative in this case. All modules should use level 1 main headings to enable seamless reuse. So, we probably should mandate the use of leveloffset.

On a related note, I don't think I understand the reasoning behind the leveloffset values suggested in the assembly template (`2` and `+2`). The explanation talks about nesting modules, which I thought is discouraged. Why put it in a template then?

rkratky (approved):

rolfedh:
> The explanation talks about nesting modules,

By nesting, do you mean something like this?

```
include:<module-file.adoc>[leveloffset=+1]
include:<module-file.adoc>[leveloffset=+2]
include:<module-file.adoc>[leveloffset=+1]
```
I've seen folks use nesting as a convenience, because they don't want to fix the H2 title in the module, not because it's necessary. I our guidelines should make a preference for [leveloffset=+1] and H1 titles clear.

rkratky:
No, I understand nesting as
```
module
     |--> module
```

> I've seen folks use nesting as a convenience, because they don't want to fix the H2 title in the module

Well, a module with H2 in the beginning is pretty much useless from mod docs perspective :(

> I think our guidelines should make a preference for [leveloffset=+1] and H1 titles clear.

+1

pwright:
@rolfedh 
there's already a note:
You can use subheadings in concept or reference modules, but not in procedure modules.

Does that address your issue? My understanding is that, assuming you've done things correctly, there is *not* a preference for [leveloffset=+1] - that's what motivated me to create the issue, I was under the impression there was that preference.

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
```diff
@@ -27,7 +27,7 @@ Use the second level heading syntax for the Prerequisites section in the assembl
 [discrete]
 == Assembly Modules
 
-List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly.
+List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute, such as `module-file.adoc[leveloffset=+1]` or `module-file.adoc[leveloffset=+2]`, to set the hierarchy of the module relative to the assembly.
```
rolfedh, line 30:
```suggestion
List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute to set the hierarchy of the module relative to the assembly. In most cases, create modules whose title is an H1 and include those modules by using `[leveloffset=+1]`. Otherwise, if necessary, you can use other leveloffsets values, to promote or demote your module content. 
```
This is one way we could recommend using `[leveloffset=+1]` and make it clear that other offsets are acceptable.

rkratky:
>  You can use subheadings in concept or reference modules, but not in procedure modules.

This sentence is weird. Every module, regardless of its type, should start with H1. And every module, regardless of its type, can use subheadings further down. I don't understand where the quoted guidance comes from, and what it is supposed to address.

> there is not a preference for [leveloffset=+1] 

I agree there should be preference to use this. That said, it's up to the writer how they structure their assemblies (incl. what level of offset they use). What's not up to the writer is how they form their modules because using anything other than H1 in the beginning of their modules makes the modules not reusable, which I consider wrong.

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
rkratky: +1 to these suggested changes. Still, I'd argue for mandating the use of H1 to start all modules. I see no reason to not use it ever.

pwright:
So, there might be a *separate* issue to explicitly state that every module starts with a H1 ?

@rkratky Are you saying that   [leveloffset=+1] is your preference? should writers be guided to create sub-assemblies and avoid using [leveloffset=+2] ?  ( asking because I think you missed the 'not' is my sentence that you quoted above)

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
sterobin: @rolfedh @rkratky I pushed a commit with a slightly tweaked version of Rolfe's suggestion, to make the examples clear and to mention specifically that offset of 2 or more is what we're saying is okay. That's the main sticking point that I want to be sure people understand is okay so they're not forcing themselves to create artificial assemblies of just two modules just to avoid a leveloffset of 2, when they think doing so is forbidden or something.

```
You can use a `leveloffset` attribute to set the hierarchy of the module relative to the assembly. In most cases, create modules whose title is an H1 and include those modules by using the format `module-file.adoc[leveloffset=+1]`. Otherwise, if necessary, you can use other `leveloffset` values, such as `module-file.adoc[leveloffset=+2]`, to modify the structure of your module content.
```

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
```diff
@@ -27,7 +27,7 @@ Use the second level heading syntax for the Prerequisites section in the assembl
 [discrete]
 == Assembly Modules
 
-List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly.
+List link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include files] to include the required modules. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. You can use a `leveloffset` attribute to set the hierarchy of the module relative to the assembly. In most cases, create modules whose title is an H1 and include those modules by using the format `module-file.adoc[leveloffset=+1]`. Otherwise, if necessary, you can use other `leveloffset` values, such as `module-file.adoc[leveloffset=+2]`, to modify the structure of your module content.
```
pwright, line 30: personally I think the mention of H1 here is confusing..  ie it's a separate issue and mentioning it here makes the text more ambiguous.

pwright (changes requested):

rolfedh:
> Does that address your issue? My understanding is that, assuming you've done things correctly, there is _not_ a preference for [leveloffset=+1] - that's what motivated me to create the issue, I was under the impression there was that preference.

It would probably be better to discuss this in one of the regular Mod Docs meetings that @emmurphy1 hosts. Please contact her for more information.

I see how the mod docs reference guide (MDRG) could be more clear on this issue. 

- It recommends starting modules with H1.
- It [shows using leveloffset to promote/demote nested modules](https://raw.githubusercontent.com/redhat-documentation/modular-docs/master/modular-docs-manual/files/TEMPLATE_ASSEMBLY_a-collection-of-modules.adoc).
```
////
[leveloffset=+1] ensures that when a module title is a level 1 heading (= Title), the heading will be interpreted as a level-2 heading (== Title) in the assembly. Use [leveloffset=+2] and [leveloffset=+3] to nest modules in an assembly.
////

include::modules/TEMPLATE_PROCEDURE_doing_one_procedure.adoc[leveloffset=+2]
include::modules/TEMPLATE_PROCEDURE_reference-material.adoc[leveloffset=2]
```

So you're right, the guide shouldn't express a preference for [leveloffset=+1]. 

The poorly articulated behavior we're trying to discourage is folks starting modules with H-whatever and then using leveloffset values to compensate for that it after the fact.

I'll revise my suggestion, above to reflect this better understanding of the issue.

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
rolfedh: > personally I think the mention of H1 here is confusing.. ie it's a separate issue and mentioning it here makes the text more ambiguous.

I don't think it is separate. It is integral to the notion of using leveloffset attributes. I'll see how I can clear up the ambiguity.

rolfedh: Revising now. Please hold off on more comments until I'm done.

rolfedh: Actually, I think Stetson's current revision encompasses all our concerns:
- It discourages "bad behavior" such as H-whatever + leveloffset=whatever
- It allows using leveloffset values to demote/promote module content.

Side note: I think we should avoid the term "nesting" in this discussion because that describes including something **within** something else, like [3.2.5. Nesting Assemblies in Assemblies](https://redhat-documentation.github.io/modular-docs/#nesting-assemblies). In that sense, we're not really talking about nesting modules within modules here. Here we're talking about using leveloffset to demote/promote modules relative to other modules in an assembly. Nesting modules within modules is a controversy we don't want to get into.

rolfedh:
Thanks for the healthy discussion, folks. I need to step away from this conversation to focus on producing content for an upcoming release two weeks away. My suggestions are just that.

rolfedh:
I think Stetson's current revision encompasses all our concerns:

- It discourages "bad behavior" such as H-whatever + leveloffset=whatever.
- It mentions using leveloffset values to demote/promote module content.

rkratky:
@pwright 

> Are you saying that [leveloffset=+1] is your preference?

Yes. In my view, this should be the default. I.e. all modules using H1 in the beginning, and assemblies including such modules using leveloffset `+1`.

> should writers be guided to create sub-assemblies

No.

> and avoid using [leveloffset=+2]

Yes.

> I think you missed the 'not' is my sentence that you quoted above

You mean in _not in procedure modules_? If that's what you mean, then no -- I don't see a reason to not use subheadings (provided the module starts with H1).

pwright:
@rkratky I was confused by:
image

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
emmurphy1: I agree with @pwright . Every module and assembly title must be an H1 so  no need to mention it here. @rkratky you cannot use H2s in procedure modules and the only H1 is the title. You can use H2s in concept and reference modules.

rkratky:
@pwright, my bad. What I meant was that I was in agreement with the assertion that the lack of preference for leveloffset=+1 was a mistake/oversight. I.e. it was my "+1" to adding the leveloffset recommendation.

sterobin:
@pwright @rkratky So are we good here? Been quite a loop and if we're now good I'd like to resolve some of these and then petition the rest of the approvals. Let me know. Thanks.

pwright (commented):
lgtm

pwright (approved):

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
emmurphy1: Hey folks. I'm sorry, but I find that updating confusing. Please consider this, which I think gets to the heart of the problem:

To include modules in an assembly, use the Asciidoc  link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include directive]. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. Use the`leveloffset` attribute to set the hierarchy of the module relative to the assembly, for example:
[source]
----
file1.adoc[leveloffset=+1]
file2.adoc[leveloffset=+2]
file3.adoc[leveloffset=+3]
----

NOTE: All module and assembly titles must use the H1 heading designator.

emmurphy1: I have no idea why the text is bolded, but you get the idea.

sterobin: Agreed with @emmurphy1 and @pwright on this. So @rolfedh and @rkratky , how would you feel about not getting into H1 to avoid the distraction from the point here?

pwright: thanks @emmurphy1 I prefer your wording

sterobin: +1 @rolfedh @rkratky You good here?

rolfedh: @sterobin All good. I trust your judgement!

emmurphy1 (approved):

Preeticp (approved):
/lgtm

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
sterobin: @emmurphy1 Done

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
```diff
@@ -27,7 +27,16 @@ Use the second level heading syntax for the Prerequisites section in the assembl
...
+To include modules in an assembly, use the Asciidoc  link:http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files[include directive]. Use any combination of concept, procedure, and reference modules that fulfills the purpose of the assembly. Use the `leveloffset` attribute to set the hierarchy of the module relative to the assembly, as shown in the following example:
+
+.Level offset for module files
+----
+file1.adoc[leveloffset=+1]
+file2.adoc[leveloffset=+2]
+file3.adoc[leveloffset=+3]
+----
+
+NOTE: All module and assembly titles must use the H1 heading designation (`=`).
```
sterobin, line 39: @emmurphy1 , @rkratky , @fbolton , @Preeticp , @mikemckiernan @yzimmerm , This was approved back in May, but we've had some shuffling, a suggestion from Emily which I've now incorporated, and a lot of time since then, so resurfacing once more. If you're all still okay with this addition, please approve and/or comment accordingly and we'll merge from there. So sorry to leave this in the dust. Fell off my radar entirely. Thanks!

mikemckiernan: I do not feel strongly, but I don't think it needs to be a note.  I know this is minutiae, but I'm also wincing at the parens because the ISG suggests avoiding parens.  Maybe use "...heading designation, `=`."

mikemckiernan (approved):
nits, but crystal clear either way.

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
sterobin: @mikemckiernan Thanks. I changed to this. That work?

>All module and assembly titles must use the H1 heading designation, such as `= My heading`.

Preeticp (approved):
LGTM

modular-docs-manual/content/topics/module_guidelines-assembly.adoc
mikemckiernan: Looks good. Thanks.

emmurphy1:
Thanks @sterobin . The changes LGTM. @rkratky, @fbolton, @yzimmerm I am going to merge these changes on Thursday if we don't hear from you.

fbolton (approved):
LGTM!

yzimmerm (approved):

Issue #179, opened 2022-01-27
Add [] to the snippet example file.


emmurphy1:
@RichardHoch reported:

I think there's a slight but important omission in the Modular Documentation Reference Guide's discussion of snippets:

https://redhat-documentation.github.io/modular-docs/#using-text-snippets

The problem is: in the include statement shown, there are no square brackets after the snippet file name. That is, the example should be `include::snippets/beta-note.adoc[]`   and not `include::snippets/beta-note.adoc` as the text currently reads.

Without those brackets, the document will build in bccutil as if everything is fine, [but the content of the snippet file is not included].

So you might think your document is OK since there is no error message in bccutil (that might be an issue, too) but it isn't.

Please check this out.  From experimenting with include statements, I see they always must be followed by square brackets -- we're used to this since that's where we insert the level offset information. Nonetheless, it might be worth mentioning that when you include a snippet you have to include the square brackets, even though it's not a module, so there is no level offset.

mjahoda:
@emmurphy1 This is a good catch. The square brackets are mandatory parts of the include syntax:
https://docs.asciidoctor.org/asciidoc/latest/directives/include/

sterobin:
Fixed in PR #180. Awaiting PR approval to merge fix.

PR #180, opened 2022-01-27
Fix bug in snippet example

sterobin:
Fixes issue #179


emmurphy1 (approved):

sterobin:
Agreed with @emmurphy1 in chat that with how trivial and definitive this is, merging with just her and my approvals.

PR #185, opened 2022-03-24
issue-184 Removed the abstract tag from all templates

emmurphy1:


sterobin (approved):
@emmurphy1 I've approved, but just pointing out that my understanding is that although the abstract tag indeed isn't required by tooling, which will pull the first paragraph as the abstract, it's a failsafe to ensure an abstract is pulled in the event that a writer doesn't include spacing between the header and the first paragraph. I believe both pv2 and Jupiter rely on the line break following the header and following the first paragraph are the delimiters for the abstract. Should anyone or anything (auto-generated content) disregard the line break(s), then the abstract tag would be the failsafe in that case. I could be wrong, but in case this comes up after we make this change, I wanted us to be aware. 

I do agree though to ditch the tag regardless for the sake of vast majority of successful cases.

modular-docs-manual/files/TEMPLATE_REFERENCE_reference-material.adoc
```diff
@@ -20,15 +20,16 @@ The ID is an anchor that links to the module. Avoid changing it after the module
 [id="ref-reference-material_{context}"]
 
 ////
-The `context` attribute enables module reuse. Every module ID includes {context}, which ensures that the module has a unique ID even if it is reused multiple times in a guide
+The `context` attribute enables module reuse. Every module ID includes {context}, which ensures that the module has a unique ID even if it is reused multiple times in a guide.
```
mjahoda, line 23: I would consider replacing the "guide" term with something already-defined or unequivocal - assembly of assemblies, an assembly that includes other assemblies, ...

modular-docs-manual/files/TEMPLATE_REFERENCE_reference-material.adoc
```diff
@@ -20,15 +20,16 @@ The ID is an anchor that links to the module. Avoid changing it after the module
...
 ////
 
 = Reference material
 ////
 In the title of a reference module, include nouns that are used in the body text. For example, "Keyboard shortcuts for ___" or "Command options for ___." This helps readers and search engines find the information quickly.
+
+Be sure to include a line break between the title and the module introduction.
```
mjahoda, line 30: OT: Do we have an instruction for adding an empty line between every two includes somewhere? (this can prevent peculiar and hard-to-debug problems)

vsodhi:
LGTM.

modular-docs-manual/files/TEMPLATE_REFERENCE_reference-material.adoc
emmurphy1: Yes, in the assembly template.

Issue #200, opened 2023-01-30
Creating modules|Procedure modules contains unclear parts


mjahoda:
https://redhat-documentation.github.io/modular-docs/#con-creating-procedure-modules_writing-mod-docs

> Procedure modules can also contain prerequisites, verification steps, and additional resources or next steps.

Does the previous sentence mean that we can have either additional resources or next steps, but not both?

If yes, then I am suggesting to allow both.

Moreover, the order of those parts is also not clear. For example, can I place next steps right after the procedure steps (when it makes sense in a particular procedure)?

emmurphy1:
@mjahoda the 'or' in that sentence means either one or the other. However, I won't object to changing it if the board agrees:
Procedure modules can also contain prerequisites, verification steps, additional resources, and next steps.

If we decide on this change, I think Next Steps should be last and we should modify the example in the template to illustrate this. I am interested to hear what others think.

emmurphy1:
@mjahoda the change that you made in the pull request (https://github.com/redhat-documentation/modular-docs/pull/202/files) is different from what I understood this request to be. You have introduced a new element to the procedure. I think we need to discuss this further before proceeding with the change.

mjahoda:
@emmurphy1 Unfortunately, I noticed your comment after I merged the PR #202. I hope that because it does not introduce any strict rule (rather the other way round), there is no harm to keep it merged until we discuss it further.

emmurphy1:
@mjahoda that's OK. @ncbaratta reverted it. Leaving it there would create confusion. I'll start a discussion on the Slack channel. I had it on the agenda to discuss at the last meeting but you were there.

PR #202, opened 2023-04-12
Clarify Next steps

mjahoda:
https://github.com/redhat-documentation/modular-docs/issues/200

### DIFF
--- a/modular-docs-manual/content/topics/creating_procedure_modules.adoc
+++ b/modular-docs-manual/content/topics/creating_procedure_modules.adoc
@@ -42,7 +42,7 @@ This section is optional. The additional resources list links to other material 
 Focus on relevant resources that might interest the user. Do not list resources for completeness. If a resource applies to all of the modules in a user story, consider listing the resource in the Additional resources section of the assembly file. If you do this, consider including the resource in the procedure module as a comment.
 
 .Procedure next steps
-This section is optional. You can use it for pointing the reader to instructions covering tasks that might follow after the current procedure. Place this section after the Additional resources section with the *Next steps* heading. Alternatively, if the user should perform the next steps before the verification steps, place the section in between the Procedure section and the Verification steps section.
+This section is optional. You can use it for pointing the reader to tasks that might follow after the current procedure. Place this section after the *Additional resources* section and use the *Next steps* heading. Alternatively, if the user needs to perform the next steps before the verification steps, place the section between the *Procedure* section and the *Verification* section.
 
 == Additional resources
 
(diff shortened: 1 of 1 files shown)


modular-docs-manual/content/topics/creating_procedure_modules.adoc
```diff
@@ -41,6 +41,9 @@ This section is optional. The additional resources list links to other material
 
 Focus on relevant resources that might interest the user. Do not list resources for completeness. If a resource applies to all of the modules in a user story, consider listing the resource in the Additional resources section of the assembly file. If you do this, consider including the resource in the procedure module as a comment.
 
+.Procedure next steps
+This section is optional. You can use it for pointing the reader to instructions covering tasks that might follow after the current procedure. Place this section after the Additional resources section with the *Next steps* heading. Alternatively, if the user should perform the next steps before the verification steps, place the section in between the Procedure section and the Verification steps section.
```
thatdocslady, line 45:
```suggestion
This section is optional. You can use it for pointing the reader to tasks that might follow after the current procedure. Place this section after the *Additional resources* section and use the *Next steps* heading. Alternatively, if the user needs to perform the next steps before the verification steps, place the section between the *Procedure* section and the *Verification* section.
```

thatdocslady (commented):
Bunch of nit picks (sorry, docs people be docs people ;). Feel free to implement or override as needed.

modular-docs-manual/content/topics/creating_procedure_modules.adoc
emmurphy1: Suggest changing 'pointing' to 'directing'. I really don't think we should add a Next steps section before the verification. If the 'next steps' are needed to complete the procedure then they should be part of the procedure.

mjahoda:
> Bunch of nit picks (sorry, docs people be docs people ;). Feel free to implement or override as needed.

Thank you, nice improvements. I've just accepted your patch, and I'll just replace "needs to" (because IBM Style discourages its usage similarly as my "should").

modular-docs-manual/content/topics/creating_procedure_modules.adoc
mjahoda: > Suggest changing 'pointing' to 'directing'. 

Changed, thank you.

> I really don't think we should add a Next steps section before the verification. If the 'next steps' are needed to complete the procedure then they should be part of the procedure.

However, an example of why we should explicitly allow such order was one of the reasons for filing this issue. Marc Muehlfeld and I believe that in cases such as [1] is this the best order.

[1] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/securing_networks/creating-and-managing-tls-keys-and-certificates_securing-networks#creating-a-private-ca-using-openssl_creating-and-managing-tls-keys-and-certificates

ncbaratta (approved):
lgtm

Issue #223, opened 2024-02-08
Add a one-line comment to the Procedure template to avoid rendering issues


Jenny-Anne:
Add a one-line comment to the Procedure template to avoid rendering issues: 

"Be sure to include a line break between the title and the module introduction."

This line is included in the other templates.
Follows on from the removal of the `[role="_abstract"]` tag in #221.

PR #224, opened 2024-02-08
Adding one line comment to procedure template

Jenny-Anne:
Addresses #223

### DIFF
--- a/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
+++ b/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
@@ -18,6 +18,8 @@ The ID is an anchor that links to the module. Avoid changing it after the module
 = Doing procedure A
 ////
 Start the title of a procedure module with a gerund, such as Creating, Installing, or Deploying.
+
+Be sure to include a line break between the title and the module introduction.
 ////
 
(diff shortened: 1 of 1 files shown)


modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
```diff
@@ -18,6 +18,8 @@ The ID is an anchor that links to the module. Avoid changing it after the module
 = Doing procedure A
 ////
 Start the title of a procedure module with a gerund, such as Creating, Installing, or Deploying.
+
+Be sure to include a line break between the title and the module introduction.
```
rolfedh, line 22: "Include" can come across as "add", which isn't needed in this case.
How about "maintain" or something similar? For example:
```suggestion
Maintain the line break between the title and module introduction.
```

rolfedh (approved):
LGTM. I've offered a minor suggestion.

modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
Jenny-Anne: Thanks Rolfe, I used the same line that is already in the other templates for consistency. So  I guess it would be best to leave the line as it is or update the same line in the other templates too.

emmurphy1 (approved):

IanFrangs (approved):
LGTM

neal-timpe (approved):

PR #242, opened 2025-05-26
Added a complete list of supported block titles

jhradilek:
This is the definitive list of titles that are recognized during the DITA conversion and mapped to the corresponding `<prereq>`, `<step>`/`<steps-unordered>`, `<result>`, `<tasktroubleshooting>`, `<postreq>`, and `<related-links>` elements. It also matches the [corresponding vale rule](https://github.com/jhradilek/asciidoctor-dita-vale) that can be used to report unsupported block titles in procedures.

I have also removed the mention that these need to be in this order because while it makes sense and DITA insists on this order, the conversion tooling actually maps them correctly regardless in what order they appear. This re-shuffling is necessary because any contents preceding the first block title in AsciiDoc is mapped to `<context>` and while in our documentation it comes first, in DITA it appears after the `<prereq>` element.

### DIFF
--- a/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
+++ b/modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
@@ -41,7 +41,18 @@ Write a short introductory paragraph that provides an overview of the module. Th
 NOTE: You can add text, tables, code examples, images, and other items to a step. However, these items must be connected to the step with a plus sign (+). Any items under the .Procedure heading and before one of the following approved headings that are not connected to the last step with a plus sign cannot be converted to DITA.
 
 ////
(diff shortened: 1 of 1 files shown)


modular-docs-manual/files/TEMPLATE_PROCEDURE_doing-one-procedure.adoc
```diff
@@ -41,7 +41,18 @@ Write a short introductory paragraph that provides an overview of the module. Th
...
+Only the following block headings can be reliably mapped to DITA:
+
+* Prerequisites or Prerequisite
+* Procedure
+* Verification, Results, or Result
+* Troubleshooting, Troubleshooting steps, or Troubleshooting step
+* Next steps or Next step
+* Additional resources
+
+With the exception of Additional resources, these titles are only allowed in a procedure module.
```
mramendi, line 53: Maybe clarify that only one of each item is allowed?  For example, Verification and Results together is not allowed

mramendi: In my experience, the clearer the limitations, the better

emmurphy1 (commented):
LGTM. Thank @jhradilek.

neal-timpe (approved):

mjahoda (approved):
LGTM

